PR #10, opened 2023-11-24
fix: Use example command: amounts

lshoo:
`amount` should be `amounts`,
and the value should be sequence.


zoedberg:
thanks for spotting this!

PR #31, opened 2024-05-13
meta: fix rustc version 

vincenzopalazzo:
Using rustc 1.70.0 and running the following command

cargo install --debug --path .

I get the following error
```
error: failed to compile `rgb-lightning-node v0.1.0 (/home/vincent/github/rgb-lightning-node)`, intermediate artifacts can be found at `/home/vincent/github/rgb-lightning-node/target`

Caused by:
  package `clap_builder v4.5.2` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.70.0
  Either upgrade to rustc 1.74 or newer, or use
  cargo update -p clap_builder@4.5.2 --precise ver
  where `ver` is the latest version of `clap_builder` supporting rustc
  1.70.0
```
So this should fix the minimum rust version to 1.74

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "stable"
+channel = "1.74"
 components = ["clippy", "rustfmt"]
 targets = []
 profile = "minimal"


zoedberg:
Thanks for discovering this. The fix is incorrect though. `rust-toolchain.toml` tells rustup which toolchain is the preferred version, not the minimum one. To change the minimum supported rust version (MSRV) `rust-version` in the `Cargo.toml` should be used. Fixed in https://github.com/RGB-Tools/rgb-lightning-node/commit/1a4868bdd9c83cff8bc6413e00a1f743ce900cb1 (I've set 1.75 since RGB deps have that and very soon we'll update to latest RGB version), closing this

Just a note: PR text says "Signed-off" but commit is not signed

vincenzopalazzo:
Hi @zoedberg,

I did not know the version inside the cargo tolm file, thanks for showing me this.

>Just a note: PR text says "Signed-off" but the commit is not signed

In fact, the Signed-off is not a way to say that the commit is signed with the PGP, but https://www.kernel.org/doc/html/latest/process/submitting-patches.html?highlight=signed%20off#developer-s-certificate-of-origin-1-1 

Old Linux Kernel habit sorry for the confusion

Issue #38, opened 2024-09-24
Channel Opening Fails Without IP Address: Peer Info Address Not Found


bitwalt:
Hi! 
We’ve encountered an issue with channel opening when an IP address is not provided. Currently, the channel opening using just the pubkey as connection string, is only successful if the initiator of the channel is the same node that executed the /connectpeer command first.

The failure occurs with the error: Invalid peer info: cannot find the address for the provided pubkey.

zoedberg:
Thanks for reporting this, I've pushed a fix and will merge the PR as soon as the CI passes

PR #39, opened 2024-09-24
add test and fix for reverse peer address in openchannel

bitwalt:
This test fails due to #38


zoedberg:
Thanks, I've added a fix.

Closes #38

PR #44, opened 2024-12-17
add signet to supported networks

bitwalt:


zoedberg:
Thanks for your contribution! I've added a commit to include a default value for the signet indexer, otherwise the code would have panicked. Merging this

PR #47, opened 2025-04-01
Add failing test for BTC-to-asset swap with vanilla and asset channels

bitwalt:
## Description
This PR adds a failing test case to demonstrate an issue with BTC-to-asset swaps using both vanilla and asset channels. The test is intentionally failing to highlight potential issues in the swap implementation.

### Test Case Details
The test `swap_btc_to_asset_with_vanilla_and_asset_channels` attempts to:
1. Set up a BTC-to-asset swap using:
   - A vanilla channel (node1 -> node2)
   - An asset channel (node2 -> node1)
2. Execute the swap with proper liquidity
3. Verify final balances

## Actual Behavior
The test fails during the `makerexecute` operation with a "No Route Found" error, suggesting that the routing algorithm is unable to find a valid path despite having sufficient liquidity.

## Expected Behavior
The test should successfully execute the swap operation since there is sufficient liquidity in the channels

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1424,6 +1424,32 @@ async fn wait_for_usable_channels(node_address: SocketAddr, expected_num_usable_
     }
 }
 
+async fn can_forward_htlc(node_address: SocketAddr, amount_msat: u64) -> bool {
+    println!("checking if node {node_address} can forward HTLC of {amount_msat} msat");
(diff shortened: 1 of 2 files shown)


zoedberg:
Closing this since the test is flawed, the `can_forward_htlc` should be called with `qty_from + rgb_htlc_min_msat` since when swapping bitcoins for assets the bitcoin amount necessary to perform the swap is also added.

PR #57, opened 2025-05-29
add swap_reverse_same_channel test

Arshia-r-m:
Add a test case to perform a swap on same channel and reverse the swap, which fails on the second one with "No route found" error.


zoedberg:
Thanks for finding this bug and providing this test to prove it!

PR #58, opened 2025-06-02
fix : Implement self-payment prevention and duplicate payment checks …

lukegao209:
Fix: Prevent self-payments and protect against pending invoice re-payments

1. 	1.	Self-payments can cause the invoice to remain in a pending state — added a check to prevent self-payments.
2. 	2.	Re-attempting payment on a pending invoice may fail and lead to force-closed channels — added a check in send_payment to avoid this edge case.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -223,6 +223,29 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/EmptyResponse'
+  /decodeassetid:
+    post:
+      summary: Decode RGB asset ID to hexadecimal format
+      description: Convert RGB format asset ID (rgb:xxx) to hexadecimal format
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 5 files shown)


zoedberg:
@lukegao209 I've understood the cause of the force-close. When coloring the commitment transactions we use a file specific for the payment hash containing the RGB balances for that payment. Therefore having a duplicate payment was causing an error on the commitment transactions coloring procedure and the simplest fix for this is to prevent duplicate payments. This also seems reasonable since in LN it's unsafe to pay the same invoice multiple times if a previous attempt succeeded or is in pending status.

Your fix was good but to avoid calling the `get_outbound_payments` method twice it seems better to move the check inside the `add_outbound_payment` method. To accelerate the procedure I've made a commit on top of the PR where I show the bug and put you as co-author. Let me know if that is ok with you, otherwise you can amend the commit to take full ownership.

lukegao209:
will create a new pr later.

PR #63, opened 2025-06-16
Feature add state api

lukegao209:
Adding a state API can help users better understand the status of the node.


zoedberg (changes requested):
Thank you!

Requested some small changes. Please also squash into a single signed commit saying `add /nodestate API` and then rebase on top of the updated master.

Finally check lint and format with the latest stable rust

PR #71, opened 2025-09-26
add test to swap bought asset from maker

Arshia-r-m:
The test added shows an unexpected behavior of the nodes after the 2nd node swapped (bought) 2 assets with the 1st node and then tried to swapped these assets with the 1st node again.


Arshia-r-m:
Thanks for the clear explanations, the expected amounts after the last swap is wrong and I'll modify it, but the fact that the the balances aren't updated remains the same.

The test tries to replicate the exact situation I faced multiple times in my testing where the maker marks the swap as failed, taker marks it as pending and the amounts will not change.

Let me know if I could provide more details.

Arshia-r-m:
> From this I get that you expect the test to fail because of the maker failing the swap, is my understanding correct? Anyway, will do more local runs and check for this kind of behavior.
>
To be exact, from my understanding, the maker marks the swap as failed because it fails to pay to taker, so the swap it self is marked as failed, but on the taker side it is stuck in pending status and also the balances doesn't change on both end.
I'll update the test asap and push it again.

Issue #50, opened 2025-04-30
Feature Request: Support for HODL Invoices and Proxy Invoices


bitwalt:
We would like to open a discussion and propose the addition of **HODL invoices** and **Proxy invoices** support in the RGB Lightning Node (RLN). These features would enable better coordination for atomic swaps and advanced cross-protocol HTLC-based interactions (e.g. RGB <> Taproot Assets swaps).

---
#### 1. **HODL Invoice**

- `POST /hodlinvoice`: Create a HODL invoice. It will not be automatically settled when paid.
- `POST /settleinvoice`: Manually settle a previously created HODL invoice (if paid).
- `POST /cancelinvoice`: Cancel a HODL invoice, even if it's been paid.

#### 2. **Proxy Invoice**

- `POST /proxyinvoice`: Takes a Lightning invoice as input and returns a new *proxy HODL invoice* with the **same payment hash**.
  - When the proxy is paid, RLN should pay the original invoice.
  - When the original invoice is settled and the preimage is known, the proxy invoice is then settled using the same preimage.

---

### 🔗 References

- Bitcoin Ops: [[Hold Invoices](https://bitcoinops.org/en/topics/hold-invoices)](https://bitcoinops.org/en/topics/hold-invoices)  
- LND API: [[addHoldInvoice](https://lightning.engineering/api-docs/api/lnd/invoices/add-hold-invoice/)](https://lightning.engineering/api-docs/api/lnd/invoices/add-hold-invoice/)  
- LN Proxy Specification: [[lnproxy/spec](https://github.com/lnproxy/spec)](https://github.com/lnproxy/spec)  
- Related use case in RGB swaps: [[RGB TaprootAssets Atomic Swaps](https://kaleidoswap.medium.com/rgb-taprootassets-atomic-swaps-c0ac6ec8643c)](https://kaleidoswap.medium.com/rgb-taprootassets-atomic-swaps-c0ac6ec8643c)

zoedberg:
Thanks for opening this.

HODL invoices are very interesting, some time ago we also considered to use them to improve the current RGB assets swap system (as kindly suggested by @adambor).
LDK doesn't have this as an out-of-the-box feature, but it should be possible to add them using some low-level methods.

There's no ETA for this task at the moment, since we're still handling more important stuff, but we are certaintly interested in this.

adambor:
I would just further add: 
- the `POST /hodlinvoice` could take the payment hash as a parameter
[...]

Issue #24, opened 2024-03-28
RGB amounts in the intermediate node's channels weren't updated after a payment forwarded


whfuyn:
I noticed a problem while testing the rgb lightning channel.

Suppose there were three nodes, A, B, and C, and two channels, A->B, B->C.

Initially, A->B had 4500->500 rgb assets, and B->C had 20->0.

After sending 6 rgb assets from A to C through A->B->C, I expected the A->B to have 4494->506, and B->C to have 14->6.

However, when I checked in the response of `listchannel` from node B, it showed that the rgb amounts in both channels weren't updated as node A and C did.

Node A
```
{
  "channels": [
    {
      "channel_id": "bbda30a0d7a0d63bc9d7e230250c482fbce25d2951991d39e840c38c12d985f6",
      "funding_txid": "f685d9128cc340e8391d9951295de2bc2f480c2530e2d7c93bd6a0d7a030dabb",
      "peer_pubkey": "031ae15fbc060b935d6be7dbe60ff4e045f21230c29d47bca09498ad2f91a71dff",
      "peer_alias": "",
      "short_channel_id": 219902325620736,
      "ready": true,
      "capacity_sat": 100000,
      "local_balance_msat": 92604999,
      "outbound_balance_msat": 91604999,
      "inbound_balance_msat": 6395001,
      "is_usable": true,
      "public": true,
      "asset_id": "rgb:24i4tVt-DJn39YHZc-uExrSrv5K-Z11utSgd3-d6NL4c3Gr-deFPMUq",
      "asset_local_amount": 4494,
      "asset_remote_amount": 506
    }
  ]
}
```

Node B
```
{
  "channels": [
    {
      "channel_id": "ab5ec3ea5b648ab20244dd4df4ece2eb02b4b9fd9c8153c87e8819aba117eb66",
      "funding_txid": "66eb17a1ab19887ec853819cfdb9b402ebe2ecf44ddd4402b28a645beac35eab",
      "peer_pubkey": "027b4535c2bb0761b8591cc16ff585d1b73998b4f12c286e9fd6946a2dfd74ece8",
      "peer_alias": null,
      "short_channel_id": 325455441887232,
      "ready": true,
      "capacity_sat": 100000,
      "local_balance_msat": 95605999,
      "outbound_balance_msat": 94605999,
      "inbound_balance_msat": 3394001,
      "is_usable": true,
      "public": true,
[...]

Issue #20, opened 2024-01-16
Does rgb-lightning-node support sending RGB assets over lightning network yet?


whfuyn:
Hi there, I'm trying to understand the transfer process of RGB assets in rgb-lightning-node.

I notice that, in `refresh_transfer`, the transaction resolution in the transition verification (`rgb_core::validation::Validator::validate_transition`) are done by the [electrum API](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-transaction-get), but that API seems to get transactions from Bitcoin and has nothing to do with the lightning network.

I'd like to know how it verifies transitions that happens on lightning network, or if it is not supported yet.

Additionally, I wonder what the relationship between `sendpayment` and `sendasset` is. I find that `sendpayment`  pays an invoice with `rgb_contract_id` and `rgb_amount`. How does that work without the consignment logic in `sendasset`?

whfuyn:
I read through the LDK related code in `rgb-lightning-node`, and it appears to have some code injected in the underlying `rust-lightning` library. Is its functionality achieved in a similar way like the satoshis in the channel?

I am a bit of confused about `sendasset` . Why does it only broadcast the PSBT only when it's a donation? For the receiver to validate the consignment, that PSBT must be published, right? Otherwise the validation status in the receiver side will be `UnresolvedTransactions`.

whfuyn:
Thanks for your help, @pocolocosv . 
I manage to make both `sendasset` and `sendpayment` work.

## sendasset
```bash
RGB_INVOICE=$(
    curl -X POST -H "Content-type: application/json" \
        -d "{ \"min_confirmations\": 0, \"asset_id\": \"$ASSET_ID\" }" \
        $PEER_RPC/rgbinvoice \
        | jq -r '.invoice'
)

DECODED_RGB_INVOICE=$(
    curl -X POST \
        -H 'Content-Type: application/json' \
        -d "{ \"invoice\": \"$RGB_INVOICE\" }" \
        'http://localhost:3001/decodergbinvoice'
)

RGB_INVOICE_RECIPIENT_ID=$(echo $DECODED_RGB_INVOICE | jq -r '.recipient_id')
RGB_INVOICE_TRANSPORT_ENDPOINTS=$(echo $DECODED_RGB_INVOICE | jq -r '.transport_endpoints')
RGB_INVOICE_AMOUNT=$(echo $DECODED_RGB_INVOICE | jq -r '.amount')
[...]

PR #91, opened 2026-01-19
Add HODL Invoices

txalkan:
This PR adds support for HODL invoices to the RGB Lightning Node per the #50 issue.

The main functional addition is the ability to create and manage HODL invoices, requiring updates across the API, core logic, error handling, persistence layer, and a new test suite. Incoming HTLCs are held and only settled or cancelled explicitly.

- Primary API additions include: `/invoice/hodl`, `/invoice/settle`, and `/invoice/cancel`.

Other adjustments to workflows and documentation ensure the feature is well integrated.

- `src/routes.rs` added new routes to create HODL invoices, settle or cancel them. It includes request validation and JSON responses that comply with the updated OpenAPI spec. Wired the new routes into `src/main.rs`. 
- `src/ldk.rs` upgraded how relevant events are processed. This includes new handlers for holding and settling payments, and updates to invoice retrieval and status checks.
- `src/test/hodl_invoice.rs` is a comprehensive test suite verifying HODL invoice creation, payment flows, and settlement/cancellation scenarios.

More information and diagrams to support advanced flows such as submarine swaps are documented in: [feat_hodl_invoice_v0.1.pdf](https://github.com/user-attachments/files/24718935/feat_hodl_invoice_v0.1.pdf).

A working submarine swap PoC (from Signet to Regtest) is available at https://github.com/UTEXO-Protocol/thunder-swap.


zoedberg (changes requested):
Thank you!

This is just a first review, will do a more detailed one after these changes are addressed:
- revert changes to the github workflows
- keep a single signed commit called `add support for HODL invoices`
- drop nix/flake related files
- drop changes to `regtest.sh` (shouldn't be necessary) or explain why they are necessary
- rename APIs as requested in the initial google doc shared at the beginning of the task
- please keep alphabetical order (in README, openapi.yaml, main.rs and routes.rs)
- drop HODL Invoices explanation from the README, let's keep API documentation in the openapi.yaml (consise doc is preferred)
- in openapi.yaml use the same syntax we used for other objects (in `required` for example)
- drop docstrings from route methods (all other methods do not have it) and keep doc in openapi.yaml
- in `src/test/hodl_invoice.rs` move imports to `src/test/mod.rs`
- in `src/test/hodl_invoice.rs` please cleanup the code, keep the API calls logic in the mod.rs file (as we did for all other tests), avoid methods that are called only once and repeating utility methods that are already defined in the mod.rs file
- in `src/utils.rs` keep methods only if used more than once

txalkan:
> * drop changes to `regtest.sh` (shouldn't be necessary) or explain why they are necessary

It waits for the Bitcoin RPC to be fully ready, preventing race conditions previously observed where Electrs or other services started too early and failed intermittently.

The rest I believe has been addressed, let me know.

txalkan:
> @txalkan Could you point me to the items that are not fully ordered please?

- https://github.com/RGB-Tools/rgb-lightning-node/blob/75f9ac50b0ec5c595fa3dcab34f91eb7e5f74d35/src/routes.rs#L2189
- https://github.com/RGB-Tools/rgb-lightning-node/blob/75f9ac50b0ec5c595fa3dcab34f91eb7e5f74d35/src/routes.rs#L2332

Could be nicer to order `mod.rs` as well, so it’s clearer where new functions should be added.

src/routes.rs
```diff
@@ -758,6 +786,22 @@ pub(crate) struct LNInvoiceResponse {
...
 }
 
+#[derive(Deserialize, Serialize)]
+pub(crate) struct InvoiceHodlRequest {
+    pub(crate) amt_msat: Option<u64>,
+    pub(crate) expiry_sec: u32,
+    pub(crate) asset_id: Option<String>,
+    pub(crate) asset_amount: Option<u64>,
+    pub(crate) payment_hash: String,
+    pub(crate) external_ref: Option<String>,
```
zoedberg: I don't think we need the `external_ref` field, please drop it or give a reason for why we should keep it.

zoedberg (changes requested):
Here another small review, will do a deeper one later on, please first address the requested changes

src/utils.rs
```diff
@@ -445,3 +458,41 @@ pub(crate) fn get_route(
...
+        preimage_vec
+            .try_into()
+            .map_err(|_| APIError::InvalidPaymentPreimage)?,
+    );
+    let computed_hash = PaymentHash(Sha256::hash(&preimage.0).to_byte_array());
+    if computed_hash != *payment_hash {
+        return Err(APIError::InvalidPaymentPreimage);
+    }
+    Ok(preimage)
+}
```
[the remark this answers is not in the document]
txalkan: `validate_and_parse_payment_hash` already follows the same validation semantics and has additional benefits, such as improved error handling. It is used multiple times, and also in the upcoming submarine swaps API. I forgot to call it from `get_payment_preimage`, but I'll refactor it to use this helper too.

src/routes.rs
```diff
@@ -3668,6 +3891,59 @@ pub(crate) async fn send_rgb(
...
+                existing.status,
+                HTLCStatus::Pending | HTLCStatus::Cancelled | HTLCStatus::Failed
+            ) {
+                return Err(APIError::InvoiceNotClaimable);
+            }
+        }
+
+        let _claimable = unlocked_state.mark_claimable_settling(&payment_hash, metadata.expiry)?;
+
+        unlocked_state.channel_manager.claim_funds(preimage);
```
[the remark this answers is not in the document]
txalkan: `claimable` and `settling` represent two different phases, and both are intentional.

- `claimable`: we received `Event::PaymentClaimable` for a HODL invoice, so the HTLC is held and awaiting user action.
- `settling`: `/settlehodlinvoice` was accepted and `claim_funds(...)` was requested, but `Event::PaymentClaimed` has not been processed yet.

`claim_funds(...)` initiates claim handling, but the terminal state transition is event-driven and asynchronous. That creates a TOCTOU window between “checked as claimable” and “claim confirmed.” We set `settling=true` before `claim_funds(...)` to close that window and prevent concurrent conflicting operations (`cancel`/second `settle`) on the same HTLC.

When `Event::PaymentClaimed` arrives, we mark the payment `Succeeded` and remove the claimable record.

src/utils.rs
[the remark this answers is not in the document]
zoedberg: your version:
```rust
if payment_hash_str.is_empty() {
    return Err(APIError::InvalidPaymentHash("missing payment_hash".into()));
}
let hash_vec = hex_str_to_vec(payment_hash_str)
    .ok_or_else(|| APIError::InvalidPaymentHash(payment_hash_str.to_string()))?;
if hash_vec.len() != 32 {
    return Err(APIError::InvalidPaymentHash(payment_hash_str.to_string()));
}
let hash_bytes: [u8; 32] = hash_vec
    .try_into()
    .map_err(|_| APIError::InvalidPaymentHash(payment_hash_str.to_string()))?;
Ok(PaymentHash(hash_bytes))
```
my version:
```rust
let payment_hash_vec = hex_str_to_vec(&payload.payment_hash);
if payment_hash_vec.is_none() || payment_hash_vec.as_ref().unwrap().len() != 32 {
    return Err(APIError::InvalidPaymentHash(payload.payment_hash));
}
Ok(PaymentHash(payment_hash_vec.unwrap().try_into().unwrap()))
```

Error handling doesn't seem improved to me. Checking the string is empty is redundant, you already check the length is 32. Also `hash_vec.try_into()` cannot fail since you already checked the vector has length 32. Therefore I fail to see the improvements you mention.

zoedberg (changes requested):
I answered to the pending conversations, please also rebase this PR on top of the updated master

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,608 @@
...
+        node2_addr,
+        payment_hash_hex,
+        StatusCode::NOT_FOUND,
+        "No claimable HTLC found for this invoice",
+        "InvoiceNotClaimable",
+    )
+    .await;
+}
+
+async fn setup_two_nodes_with_asset_channel(
```
zoedberg: as previously said, please remove unnecessary utilities and keep the tests style consistent with existing tests

zoedberg (changes requested):
> Important clarification on HODL lifecycle: `PaymentClaimable` does not always reveal the preimage. In the hash-based HODL flow (invoice created from a provided payment hash), `payment_preimage` can be `None`, so we cannot rely on that event to store/use the preimage.

Oh correct, sorry I got confused during the review.

Left some other changes requests. Generally speaking please try to keep the diff as small as possible. Cosmetic changes make the review harder.

Also I see you didn't change `Payment`, therefore when listing payments it's impossible to distinguish hodl invoices from auto claim ones. I propose to replace the `inbound` bool with an enum like
```rust
enum PaymentType {
    Outbound,
    InboundAutoClaim,
    InboundHodl,
}
```

zoedberg (changes requested):
I'm sorry I still haven't found the time to re-review this and answer to the pending comments. In the meatime though please add a test that uses HODL invoices to perform a swap. This test should show an issue reported in https://github.com/RGB-Tools/rust-lightning/issues/20 and https://github.com/RGB-Tools/rust-lightning/issues/24

txalkan:
> issue reported in [RGB-Tools/rust-lightning#20](https://github.com/RGB-Tools/rust-lightning/issues/20) and [RGB-Tools/rust-lightning#24](https://github.com/RGB-Tools/rust-lightning/issues/24)

It seems like @free-free-6 was right to point out this issue. I've been investigating it here (https://github.com/txalkan/rgb-lightning-node/commit/0c97f7e302cb68861fc21f4df7e9e7eaa295e24b), and BTC-only payouts fail when an RGB pay-in with the same hash is already present. This does not affect the RGB/USDT use case (e.g., async payments) where the outbound leg is RGB-backed. To support the latter, I made `get_payment` lookup type-aware, so inbound and outbound records sharing the same hash are no longer conflated.

src/test/mod.rs
```diff
@@ -991,15 +1010,57 @@ async fn ln_invoice(
...
+    .await
+}
+
+async fn ln_invoice_hodl(
+    node_address: SocketAddr,
+    amt_msat: Option<u64>,
+    asset_id: Option<&str>,
+    asset_amount: Option<u64>,
+    expiry_sec: u32,
+    payment_hash: Option<String>,
```
zoedberg: HODL invoices are the ones that have they payment_has set, so the parameter here must be non optional

src/test/mod.rs
```diff
@@ -991,15 +1010,57 @@ async fn ln_invoice(
...
+}
+
+async fn ln_invoice_with_type(
+    node_address: SocketAddr,
+    amt_msat: Option<u64>,
+    asset_id: Option<&str>,
+    asset_amount: Option<u64>,
+    expiry_sec: u32,
+    payment_hash: Option<String>,
+    invoice_type: InvoiceType,
```
zoedberg: no need for this `invoice_type` parameter here, just log the optional payment hash and it will be clear if it's HODL or not

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,632 @@
...
+    let res = reqwest::Client::new()
+        .post(format!("http://{node_address}/claimhodlinvoice"))
+        .json(&payload)
+        .send()
+        .await
+        .unwrap();
+    check_response_is_nok(res, expected_status, expected_message, expected_name).await
+}
+
+async fn run_auto_claim_invoice_regression_case(node1_addr: SocketAddr, node2_addr: SocketAddr) {
```
zoedberg: no need for this, since there are already other tests with normal invoices and payments. please remove it

src/ldk.rs
```diff
@@ -712,21 +845,123 @@ async fn handle_ldk_events(
...
+                        payment_hash,
+                        HTLCStatus::Failed,
+                        payment_preimage,
+                        payment_secret,
+                        Some(amount_msat),
+                        unlocked_state.channel_manager.get_our_node_id(),
+                    );
+                    return Ok(());
+                }
+            }
```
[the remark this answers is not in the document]
zoedberg: could you please point me to where this is documented in LDK?

src/ldk.rs
```diff
@@ -180,6 +197,75 @@ impl UnlockedAppState {
...
+        }
+        if let Some(since) = payment.claiming_since {
+            const ONE_DAY_SECS: u64 = 24 * 60 * 60;
+            if now_ts <= since.saturating_add(ONE_DAY_SECS) {
+                return false;
+            }
+        }
+        let deadline_passed = payment
+            .claim_deadline_height
+            .map(|h| current_height >= h)
```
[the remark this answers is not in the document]
zoedberg: by checking LDK you can see that `claim_deadline` is always set to `Some`, I think it's optional only to support old LDK versions. so please use `expect` here. I would also rename `claim_deadline_height` to `claim_deadline` so that's easier to see the connection between the 2

zoedberg (changes requested):
Please address the requested changes.

The `hodl_invoice::claim_hodl_invoice_btc_rgb` test is failing on my machine but passing on the CI, this means it's not deterministic. Please fix it. Here the fail logs:
```
thread 'test::hodl_invoice::claim_hodl_invoice_btc_rgb' (3551184) panicked at src/test/hodl_invoice.rs:473:5:
assertion failed: matches!(invoice_status(node2_addr, &invoice).await, InvoiceStatus::Pending)
```

Generally speaking tests are very messy and hard to read, could you please add some comments to separate subtests that are in the same test and briefly explain what that subtest is checking/doing?

Please also lint the code and rebase this on top of the latest master tip.

> It seems like @free-free-6 was right to point out this issue. I've been investigating it here (txalkan@0c97f7e), and BTC-only payouts fail when an RGB pay-in with the same hash is already present. This does not affect the RGB/USDT use case (e.g., async payments) where the outbound leg is RGB-backed. To support the latter, I made get_payment lookup type-aware, so inbound and outbound records sharing the same hash are no longer conflated.

Could you please add the swap test to this PR? About the fix I would like to discuss it after seeing the test reproducing the reported issue.

zoedberg:
Hi @txalkan, quick ping on this PR. Could you share what is currently blocking progress here?

From the latest comments, it looks like the remaining work is around the nondeterministic HODL invoice test, adding the swap test, test cleanup, linting, and rebasing. If there is a design question or review point that needs clarification, let us know so we can unblock it.

txalkan:
> Hi @txalkan, quick ping on this PR. Could you share what is currently blocking progress here?
> 
> From the latest comments, it looks like the remaining work is around the nondeterministic HODL invoice test, adding the swap test, test cleanup, linting, and rebasing. If there is a design question or review point that needs clarification, let us know so we can unblock it.

Hi @zoedberg, all good! I've just been busy with virtual channels and async payments, but I'll get back to this as soon as possible.

txalkan:
Hi @zoedberg,

- Here's the LDK reference regarding the expiry check: [channelmanager.rs](https://github.com/RGB-Tools/rust-lightning/blob/rgb/lightning/src/ln/channelmanager.rs#L13520-L13522)
- I added a new test that reproduces the issue in which inbound RGB metadata obstructs an outbound BTC payment when sharing the same payment hash.
- While working on abandoning a virtual channel, I noticed that RGB metadata remained after an invoice was cancelled or failed. `clear_rgb_payment_pending` is intended to remove those stale pending artifacts. For more context, here's the issue: https://github.com/UTEXO-Protocol/rgb-lightning-node/issues/21

src/ldk.rs
```diff
@@ -1054,21 +1179,94 @@ async fn handle_ldk_events(
...
+                        .get_inbound_payments()
+                        .payments
+                        .get(&payment_hash)
+                        .cloned()
+                    else {
+                        tracing::error!(
+                            "Missing inbound payment state for claimable payment {:?}",
+                            payment_hash
+                        );
+                        return Err(ReplayEvent());
```
zoedberg: this is a permanent error and treating it as transient will make the node stop processing every subsequent event. moreover I think this condition should never happen, so I think that we can use `expect` here

src/routes.rs
```diff
@@ -1716,6 +1793,74 @@ pub(crate) async fn check_proxy_endpoint(
...
+                    return Err(APIError::InvoiceExpired);
+                }
+            }
+
+            existing_payment_mut.status = HTLCStatus::Claiming;
+            existing_payment_mut.updated_at = now_ts;
+            unlocked_state.save_inbound_payments(inbound);
+        }
+
+        unlocked_state.channel_manager.claim_funds(preimage);
```
zoedberg, line 1855: this is not guaranteed to succeed, this means with the current code a HODL transfer could stay stuck in `Claiming` status. I think we should handle the `HTLCHandlingFailed` event

src/routes.rs
```diff
@@ -1716,6 +1793,74 @@ pub(crate) async fn check_proxy_endpoint(
...
+) -> Result<Json<ClaimHodlInvoiceResponse>, APIError> {
+    no_cancel(async move {
+        let guard = state.check_unlocked().await?;
+        let unlocked_state = guard.as_ref().unwrap();
+
+        let payment_hash = validate_and_parse_payment_hash(&payload.payment_hash)?;
+        let preimage =
+            validate_and_parse_payment_preimage(&payload.payment_preimage, &payment_hash)?;
+
+        {
```
zoedberg: why are you putting this code into brackets? they don't seem necessary

src/routes.rs
```diff
@@ -1716,6 +1793,74 @@ pub(crate) async fn check_proxy_endpoint(
...
+            if !matches!(existing_payment_mut.invoice_type, Some(InvoiceType::Hodl)) {
+                return Err(APIError::InvoiceNotHodl);
+            }
+
+            match existing_payment_mut.status {
+                HTLCStatus::Succeeded => {
+                    let computed_hash = PaymentHash(Sha256::hash(&preimage.0).to_byte_array());
+                    if computed_hash != payment_hash {
+                        return Err(APIError::InvalidPaymentPreimage);
+                    }
```
zoedberg: this seems redundant, we already checked this when calling `validate_and_parse_payment_preimage`

src/ldk.rs
```diff
@@ -623,6 +715,39 @@ fn find_and_update_rgb_chan_amt(ldk_data_dir: &Path, payment_hash: &PaymentHash,
...
+) {
+    let payment_hash_str = hex_str(&payment_hash.0);
+    let payment_path = get_rgb_payment_info_path(payment_hash, ldk_data_dir, inbound);
+    let extension = payment_path
+        .extension()
+        .expect("RGB payment info path has an extension")
+        .to_string_lossy();
+    let pending_extension = format!(".{extension}_pending");
+
+    let Ok(entries) = fs::read_dir(ldk_data_dir) else {
```
zoedberg: I don't think we need to scan all files here, the pending file should always be `{payment_hash}.{inbound,outbound}_pending`

src/test/mod.rs
```diff
@@ -196,6 +201,10 @@ async fn check_response_is_ok(res: Response) -> Response {
     res
 }
 
+async fn _check_response_is_ok(res: Response) -> Response {
+    check_response_is_ok(res).await
+}
+
```
zoedberg: this can be dropped, it's useless

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,750 @@
...
+        if (OffsetDateTime::now_utc() - t_0).as_seconds_f32() > 20.0 {
+            return Err(APIError::Unexpected(format!(
+                "inbound entry for {payment_hash} did not reach state {expected_status:?}"
+            )));
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+    }
+}
+
+async fn wait_for_ln_payment_with_timeout(
```
zoedberg, line 249: I think this should move to `test/mod.rs`

src/routes.rs
```diff
@@ -2437,10 +2578,13 @@ pub(crate) async fn keysend(
...
                 amt_msat: Some(amt_msat),
+                claim_deadline_height: None,
                 created_at,
-                updated_at: created_at,
-                payee_pubkey: dest_pubkey,
                 expires_at: None,
+                invoice_type: None,
+                payee_pubkey: dest_pubkey,
+
+                updated_at: created_at,
```
zoedberg: moving the existing fields and the empty line seem unnecessary changes

openapi.yaml
```diff
@@ -2635,11 +2711,17 @@ components:
           type: integer
           format: int64
           example: 12345678901234567890
+    PaymentType:
+      type: string
+      enum:
+        - Outbound
+        - InboundAutoClaim
+        - InboundHodl
```
zoedberg: I think this should go after `Payment` to keep the alphabetical ordering

src/ldk.rs
```diff
@@ -1054,21 +1179,94 @@ async fn handle_ldk_events(
...
+                        payment_hash,
+                        HTLCStatus::Failed,
+                        &static_state.ldk_data_dir,
+                    );
+                    return Ok(());
+                }
+            }
+
+            if let Some(expected) = invoice.amt_msat {
+                if amount_msat < expected {
```
zoedberg, line 1225: I think here we should check also the expected rgb asset amount. this also needs a test IMO

src/ldk.rs
```diff
@@ -1054,21 +1179,94 @@ async fn handle_ldk_events(
...
             };
-            unlocked_state
-                .channel_manager
-                .claim_funds(payment_preimage.unwrap());
+
+            let now_ts = get_current_timestamp();
+            if let Some(expiry) = invoice.expires_at {
+                if now_ts >= expiry {
+                    tracing::warn!(
+                        "Received HTLC for expired invoice {payment_hash:?} (expiry {expiry})"
```
zoedberg, line 1212: I think strict expiry is important only for HODL swaps, not for every type of transfer. In a normal transfer I think we should accept a payment, even if it arrived a bit late (from my understanding the event handler doesn't send us the event if it expired a long time ago)

src/ldk.rs
```diff
@@ -1054,21 +1179,94 @@ async fn handle_ldk_events(
...
+                    unlocked_state.fail_htlc_backwards_and_update_inbound_payment(
+                        payment_hash,
+                        HTLCStatus::Failed,
+                        &static_state.ldk_data_dir,
+                    );
+                    return Ok(());
+                }
+            }
+
+            if let Some(expected) = invoice.amt_msat {
```
zoedberg, line 1224: I think in auto-claim this is already enforced by LDK, if so please check this only for HODL transfers

src/routes.rs
```diff
@@ -2916,24 +3065,43 @@ pub(crate) async fn ln_invoice(
...
+                }
+                Some(payment_hash)
+            }
+            None => None,
+        };
+
         let invoice_params = Bolt11InvoiceParameters {
             amount_msats: payload.amt_msat,
             invoice_expiry_delta_secs: Some(payload.expiry_sec),
+            payment_hash: requested_payment_hash,
```
zoedberg, line 3100: I think we should also set `min_final_cltv_expiry_delta` here (configured in the API request body), default of 6 blocks is too tight for the submarine-swap use case

src/routes.rs
```diff
@@ -1716,6 +1793,74 @@ pub(crate) async fn check_proxy_endpoint(
...
+                if current_height >= deadline_height {
+                    return Err(APIError::ClaimDeadlineExceeded);
+                }
+            }
+
+            if let Some(expiry) = existing_payment_mut.expires_at {
+                if now_ts >= expiry {
+                    return Err(APIError::InvoiceExpired);
+                }
+            }
```
zoedberg: I think this is wrong, invoice string expiration should not be enforced here, the payment already started and the CLTV deadline is what matters here

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,750 @@
...
+        &test_dir_node2,
+        ExpiryTrigger::Blocks,
+    )
+    .await;
+}
+
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn reproduce_issue_inbound_rgb_hash_reused_for_outbound_btc() {
```
zoedberg: a test that shows a failure should fail, not succeed. please make it fail, then apply a commit with the fix to show the test then passes

src/routes.rs
```diff
@@ -1716,6 +1793,74 @@ pub(crate) async fn check_proxy_endpoint(
...
+pub(crate) async fn claim_hodl_invoice(
+    State(state): State<Arc<AppState>>,
+    WithRejection(Json(payload), _): WithRejection<Json<ClaimHodlInvoiceRequest>, APIError>,
+) -> Result<Json<ClaimHodlInvoiceResponse>, APIError> {
+    no_cancel(async move {
+        let guard = state.check_unlocked().await?;
+        let unlocked_state = guard.as_ref().unwrap();
+
+        let payment_hash = validate_and_parse_payment_hash(&payload.payment_hash)?;
+        let preimage =
```
zoedberg, line 1802: I think we should save the preimage and improve the test that now is not checking the preimage is not null when calling `/getpayment`

src/ldk.rs
```diff
@@ -323,6 +338,26 @@ impl UnlockedAppState {
         Ok(())
     }
 
+    pub(crate) fn fail_htlc_backwards_and_update_inbound_payment(
```
zoedberg, line 351: why is this dropping the preimage and secret?

src/routes.rs
```diff
@@ -1675,6 +1716,42 @@ pub(crate) async fn btc_balance(
...
+            .cloned()
+            .ok_or(APIError::UnknownLNInvoice)?;
+        if !matches!(payment_info.invoice_type, Some(InvoiceType::Hodl)) {
+            return Err(APIError::InvoiceNotHodl);
+        }
+        match payment_info.status {
+            HTLCStatus::Succeeded => return Err(APIError::InvoiceAlreadyClaimed),
+            HTLCStatus::Claimable => {}
+            HTLCStatus::Claiming => return Err(APIError::InvoiceSettlingInProgress),
+            _ => return Err(APIError::InvoiceNotClaimable),
```
zoedberg: shouldn't this error be `InvoiceNotCancellable`? also, why cancel is not idempotent while claim-after-settle is? I think they should behave the same way, I think they should both fail, unless you see a reason for having idempotency in che claim-after-settle case

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,750 @@
...
+            StatusCode::CONFLICT,
+            "Invoice is already claimed",
+            "InvoiceAlreadyClaimed",
+        )
+        .await;
+    } else {
+        let status = cancel_while_claiming_res.status();
+        let body = cancel_while_claiming_res.text().await.unwrap_or_default();
+        panic!("expected 403 settling-in-progress or 409 already claimed, got {status}: {body}");
+    }
```
zoedberg, line 543: I think the test should be deterministic

src/test/hodl_invoice.rs
```diff
@@ -0,0 +1,750 @@
...
+
+    // verify that RGB balances are unchanged
+    wait_for_ln_balance(node1_addr, &asset_id, initial_ln_rgb_balance_node1).await;
+    wait_for_ln_balance(node2_addr, &asset_id, initial_ln_rgb_balance_node2).await;
+}
+
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn claim_hodl_invoice_btc_rgb() {
```
zoedberg, line 419: add a restart while `Claimable`

openapi.yaml
```diff
@@ -2389,6 +2459,12 @@ components:
             - integer
             - 'null'
           example: 42
+        payment_hash:
+          type:
+            - string
+            - 'null'
+          description: Optional. When provided, the invoice is created as HODL.
```
zoedberg: please to keep the spec style consistent avoid description on fields

src/ldk.rs
```diff
@@ -342,28 +377,77 @@ impl UnlockedAppState {
...
+                HTLCStatus::Claimable => {
+                    let claim_deadline = payment_info
+                        .claim_deadline_height
+                        .expect("claimable payment must have a claim deadline");
+                    let deadline_passed = height >= claim_deadline;
+                    let invoice_expired = payment_info
+                        .expires_at
+                        .map(|expires_at| now >= expires_at)
+                        .unwrap_or(false);
+
```
zoedberg, line 416: this is dangerous, once the HTLC is accepted the invoice string expiry is meaningless and by enforcing it here you might end up with failing the inbound leg back, after you already paid out

zoedberg (changes requested):
@txalkan please see the requested changes

zoedberg:
> While working on abandoning a virtual channel, I noticed that RGB metadata remained after an invoice was cancelled or failed. `clear_rgb_payment_pending` is intended to remove those stale pending artifacts. For more context, here's the issue: [HODL cancellation/expiry leaves stale pending markers UTEXO-Protocol/rgb-lightning-node#21](https://github.com/UTEXO-Protocol/rgb-lightning-node/issues/21)


The inbound part looks like a no-op in `clear_rgb_payment_pending`: `write_rgb_payment_info_file` is the only thing that writes a `_pending` file, and it's only called from `keysend`/`maker_execute` /`send_payment`, always with `inbound: false`. On the receiving side `color_commitment` writes `{hash}.inbound` but never a `_pending` one. Same for the two calls in `PaymentForwarded`. And the virtual-channel cleanup path from the issue doesn't exist here.

src/test/hodl_invoice.rs
txalkan: Hi @zoedberg, the fix is ready to merge in https://github.com/RGB-Tools/rust-lightning/pull/38.

PR #92, opened 2026-01-29
sea-orm integration with sqlite

dcorral:
## Summary

Replaces file-based persistence with SQLite database using sea-orm ORM.

### Changes

**New database tables:**
- `mnemonic` - encrypted mnemonic storage
- `kv_store` - LDK persistence (channel manager, payments, swaps, network graph, scorer)
- `config` - configuration key/value pairs (indexer_url, proxy_url, etc.)
- `revoked_token` - revoked auth token IDs
- `channel_peer` - channel peer pubkey/address mapping

**Migrated to database:**
- Mnemonic (was encrypted file)
- All LDK KVStore data (channel manager, inbound/outbound payments, swaps, scorer)
- Config parameters (synced to files for rust-lightning compatibility)
- Revoked tokens (was flat file)
- Channel peer data (was flat file)
- PSBTs (temporary storage during channel funding)
- Output spender transactions

**Architecture:**
- Single shared `DatabaseConnection` via `Arc`
- `SeaOrmKvStore` implements rust-lightning's `KVStoreSync` trait
- Migrations run automatically on startup
- Config stored in DB as source of truth, synced to files for rust-lightning compatibility


dcorral:
All comments have been addressed, here some comments on the sepecific points:

> I see no DB transaction usage, is this because there are no cases where they are required?

In short; no, there are no cases where they are required directly. RGB transactions are stored in rgb-lib's database and lightning payments are stored via LDK's KVStore (now in `kv_store` table). `InboundPaymentInfoStorage` and `OutboundPaymentInfoStorage` are serialized as binary blobs. LDK manages now its own persistance through `SeaOrmKvStore`. 

> Please explain the purpose of SeaOrmKvStore

SeaOrmKvStore serves the same purpose as ldk-node's `SqliteStore` ([ldk-node/src/io/sqlite_store/mod.rs](https://github.com/lightningdevkit/ldk-node/blob/5d830001c732dd1dc417fee6063eebee6e91493b/src/io/sqlite_store/mod.rs#L50)), it implements LDK's `KVStoreSync` trait to persist Lightning node state to SQLite.

By default LDK uses filesystem storage, but ldk-node moved to SQLite for better reliability. We follow the same pattern here.

Since RLN now uses sea-orm, `SeaOrmKvStore` keeps everything under a single database abstraction rather than introducing a separate persistence layer.

> Please explain or revert the changes to .gitignore

Changed to `**/target` following rgb-lib's pattern.

> Please avoid async-std since it's deprecated

Tried changing to tokio runtime but tests are not even running, there is a deeper issue involved. Let me know your thoughts on this I'm not so experienced with async rust and any help is greatly appreciated.

> Please use the latest version of sea-orm

Bumped version to 1.1.19. I see there is a new version 2.x.x but I can only find release candidates and this version introduces breaking changes on the API. I'd say staying in 1.1.19 is safer for now until a stable v2 is realeased. I'm happy to explore the upgrade to v2 when the time comes. It introduces a sync crate that could be a good fit but currently only supports rusqlite afaics. [sea-orm-sync](https://www.sea-ql.org/blog/2025-12-12-sea-orm-2.0/)

Issue #17, opened 2023-12-14
get asset balance question


cymqqqq:
my call interface here:
1.listassets, get an asset id
2.rgbinvoice, get a rgbinvoice string
3.decode rgbinvoice, get related information
4.send asset
question: I sent 10 assets, but when I checked the asset balance(getassetbalance interface), it's still the original amount(balance not changed), I don't know why. 
Such as, I sent 10, and then I checked the asset balance:
"data": {
    "settled": 100,
    "future": 100,
    "spendable": 100,
    "offchain_outbound": 0,
    "offchain_inbound": 0
  }
 nothing changed.

nicbus:
Hi, it's hard to come up with an explanation with the given data, can you please add more details? A minimal way to reproduce the issue including full commands would be best.

I suppose you're using 2 nodes, one issuing and sending the asset and the other receiving it. So for a minimal way to reproduce the issue I'd expect a list of commands and related output for both nodes.

cymqqqq:
OK full commands. for rgb-lightning-node openapi:
1./listassets
2./rgbinvoice
3./decodergbinvoice
4./sendasset
5./assetbalance
I just run one regtest node here

nicbus:
Running a single node is not a well-tested path, as for LN it typically makes sense to have at least a peer to connect to. Please try to reproduce the issue using two nodes, with the first acting as issuer and sender, and the second acting as receiver.

That said, what I meant by "A minimal way to reproduce the issue including full commands" is the complete series of commands that leads to the issue, starting from an empty node.

Command parameters are important in order to understand what's going on, so please include those. The received output can give insight as well, so include that, too.

As an example, a full command looks like:
```shell
curl -X POST -H "Content-type: application/json" \
    -d '{"password": "password"}' \
    http://localhost:3001/init
```
which can give an output like:
```json
{"mnemonic":"ready tuition nephew price struggle close satisfy legal repair pole old coconut"}
```

When reporting, please provide the full list of commands (with parameters) and related output, starting from the very beginning (/init, /unlock, /issueasset, ...) and up to the issue manifestation.

cymqqqq:
1. start a daemon:
nohup rgb-lightning-node user:password@localhost:18443 dataldk0/     --daemon-listening-port 3001 --ldk-peer-listening-port 9735 --network regtest &
[1] 162070
2. init the node:
curl -X POST -H "Content-type: application/json" \
>     -d '{
>   "password": "nodepassword"
[...]

PR #95, opened 2026-02-17
sea-orm integration + rust-lightning persitence layer

dcorral:
## Summary

Replaces file-based persistence with SQLite database using sea-orm ORM.

### Changes

**New database tables:**
- `mnemonic` - encrypted mnemonic storage
- `kv_store` - LDK persistence (channel manager, payments, swaps, network graph, scorer)
- `config` - configuration key/value pairs (indexer_url, proxy_url, etc.)
- `revoked_token` - revoked auth token IDs
- `channel_peer` - channel peer pubkey/address mapping

**Migrated to database:**
- Mnemonic (was encrypted file)
- All LDK KVStore data (channel manager, inbound/outbound payments, swaps, scorer)
- Config parameters (synced to files for rust-lightning compatibility)
- Revoked tokens (was flat file)
- Channel peer data (was flat file)
- PSBTs (temporary storage during channel funding)
- Output spender transactions

**Architecture:**
- Single shared `DatabaseConnection` via `Arc`
- `SeaOrmKvStore` implements rust-lightning's `KVStoreSync` trait
- Migrations run automatically on startup
- Config stored in DB as source of truth, synced to files for rust-lightning compatibility

**rust-lightning submodule (`961313c` → `b061470`):**
- RGB data persistence moved from filesystem to `KVStoreSync` trait across 8 files
- `rgb_utils/mod.rs`: replaced file I/O (`read_rgb_transfer_info`, `write_rgb_transfer_info`, `parse_rgb_payment_info`, `get_rgb_channel_info_path`) with KVStore namespace-based read/write operations using `RGB_PRIMARY_NS` + secondary namespaces (`channel_info`, `payment_info_inbound`, `payment_info_outbound`, `transfer_info`, etc.)
- `ChannelContext`, `ChannelManager`, `OutboundPayments`, `KeysManager`: added `rgb_kv_store: Arc<dyn KVStoreSync + Send + Sync>` field, threaded through constructors and deserialization
- `color_commitment`, `color_htlc`, `color_closing`: now take `&dyn KVStoreSync` instead of reading files directly
- `is_channel_rgb`, `get_rgb_channel_info`, `rename_rgb_files`, `filter_first_hops`, `is_payment_rgb`: switched from path-based lookups to KVStore reads
- Removed filesystem fallback — KVStore is now the sole persistence backend for RGB data

IMPORTANT!
This PR depends on https://github.com/RGB-Tools/rust-lightning/pull/17 and .submodule must be updated to point to the right RGB-Tools fork before merging


src/utils.rs
```diff
@@ -350,6 +354,28 @@ pub(crate) async fn start_daemon(args: &UserArgs) -> Result<Arc<AppState>, AppEr
...
+    // Initialize the shared database connection
+    let db_path = get_db_path(&args.storage_dir_path);
+    let connection_string = format!("sqlite:{}?mode=rwc", db_path.display());
+    let mut opt = ConnectOptions::new(connection_string);
+    // Use single connection to avoid deadlocks
+    opt.max_connections(1)
+        .min_connections(0)
+        .connect_timeout(Duration::from_secs(8))
+        .idle_timeout(Duration::from_secs(8))
+        .max_lifetime(Duration::from_secs(8));
```
zoedberg:
```suggestion
    opt.max_connections(8)
        .min_connections(1)
        .connect_timeout(Duration::from_secs(8))
        .idle_timeout(Duration::from_mins(5))
        .max_lifetime(Duration::from_hours(1));
```

src/runtime.rs
```diff
@@ -0,0 +1,25 @@
...
+
+static DB_RUNTIME: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
+    tokio::runtime::Builder::new_current_thread()
+        .thread_name("db-runtime")
+        .enable_all()
+        .build()
+        .expect("Failed to create database tokio runtime")
+});
+
+pub(crate) fn block_on<F>(future: F) -> F::Output
```
zoedberg: I think we should use the DB in an async way (see comment in the rust-lightning PR) and drop this entire file

src/utils.rs
```diff
@@ -350,6 +354,28 @@ pub(crate) async fn start_daemon(args: &UserArgs) -> Result<Arc<AppState>, AppEr
     let ldk_data_dir = args.storage_dir_path.join(LDK_DIR);
     let logger = Arc::new(FilesystemLogger::new(ldk_data_dir.clone()));
 
+    // Initialize the shared database connection
+    let db_path = get_db_path(&args.storage_dir_path);
+    let connection_string = format!("sqlite:{}?mode=rwc", db_path.display());
```
zoedberg: I don't think it's safe to use `db_path.display()`, please see how we do this in rgb-lib (`adjust_canonicalization`)

migration/src/m20250127_000001_init_db.rs
```diff
@@ -0,0 +1,166 @@
...
+            .create_table(
+                Table::create()
+                    .table(Mnemonic::Table)
+                    .if_not_exists()
+                    .col(
+                        ColumnDef::new(Mnemonic::Id)
+                            .integer()
+                            .not_null()
+                            .primary_key()
+                            .default(1),
```
zoedberg: I don't think we should set defaults for the indexes

migration/src/m20250127_000001_init_db.rs
```diff
@@ -0,0 +1,166 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(Mnemonic::Table)
+                    .if_not_exists()
+                    .col(
+                        ColumnDef::new(Mnemonic::Id)
```
zoedberg: let's call the indexes `Idx`, since that's how we did on all our other projects

migration/src/m20250127_000001_init_db.rs
```diff
@@ -0,0 +1,166 @@
...
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(Mnemonic::Table)
```
zoedberg: Can't we save the mnemonic in the `Config` table?

migration/src/m20250127_000001_init_db.rs
```diff
@@ -0,0 +1,166 @@
...
+                Table::create()
+                    .table(Config::Table)
+                    .if_not_exists()
+                    .col(
+                        ColumnDef::new(Config::Key)
+                            .string()
+                            .not_null()
+                            .primary_key(),
+                    )
+                    .col(ColumnDef::new(Config::Value).string().not_null())
```
zoedberg: Instead of saving values as strings in multiple rows I think it's better to have a single row DB table with one column per value using the appropriate type for each column (see how we did this in rgb-multisig-hub)

src/ldk.rs
```diff
@@ -111,6 +124,47 @@ pub(crate) const FEE_RATE: u64 = 7;
 pub(crate) const UTXO_SIZE_SAT: u32 = 32000;
 pub(crate) const MIN_CHANNEL_CONFIRMATIONS: u8 = 6;
 
+/// Save config to database (source of truth) and sync to file for rust-lightning compatibility.
+fn save_config_and_sync_file(
```
zoedberg: We don't need backwards compatibility so please drop all the files (mnemonic, bitcoin network etc)

src/database/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+        let result = block_on(
+            ChannelPeerEntity::find()
+                .filter(ChannelPeerColumn::Pubkey.eq(pubkey))
+                .one(self.get_connection()),
+        )?;
+
+        if let Some(peer) = result {
+            block_on(peer.delete(self.get_connection()))?;
+        }
+
```
zoedberg: I think here you can just do `delete_many().filter(...)` instead of reading and then deleting

zoedberg (changes requested):
Thanks! In addition to the requested changes please
- commit the `src/database/entities/prelude.rs` file that gets generated by sea-orm-cli
- add the migration cargo lock file to the `.gitignore` file and remove the file
- do not make changes to `src/database/entities/mod.rs` since its autogenerated by sea-orm-cli
- I know you took inspiration from rgb-lib when creating aliases like ChannelPeerActMod, but I saw this is not the best/recommended approach so please drop the aliases (see rgb-multisig-hub to see what is the best way to use the DB utility objects)
- as requested in the rust-lightning PR, please keep only essential comments

P.S this is not a final review, will do another one after these requests have been addressed

Arshia-r-m:
Hi every one, what's the status of this PR?
I think this can help us a lot since on the nodes we maintain we face a lot of file corruptions which I believe can be enhanced with this!

zoedberg:
I agree the changes in this PR are important and would like to understand if this has been paused intentionally. From what I see, the feedback I provided before hasn't been addressed yet, and there are also some merge conflicts. @dcorral  please let us know

dcorral:
Yes I needed to postpone the work on this PR, anyway I believe I'll have some time starting next week to address the requested changes and push this work forward.

src/ldk.rs
```diff
@@ -526,35 +544,62 @@ pub(crate) type OutputSweeper = ldk_sweep::OutputSweeper<
...
+        let namespace = if inbound {
+            RGB_PAYMENT_INFO_INBOUND_NS
+        } else {
+            RGB_PAYMENT_INFO_OUTBOUND_NS
+        };
+
+        if let Ok(keys) = kv_store.list(RGB_PRIMARY_NS, namespace) {
+            for key in keys {
+                // proxy keys have format `{channel_id}{payment_hash}`; bare payment_hash is skipped
+                if key.contains(&payment_hash_str) && key != payment_hash_str {
```
zoedberg:
```suggestion
                if key.ends_with(&payment_hash_str) && key.len() > payment_hash_str.len() {
```

The filter `key.contains(&payment_hash_str) && key != payment_hash_str` also matches `{payment_hash}_pending` keys, which live in the same namespaces. When that happens, `channel_id_str` becomes `_pending` and `update_rgb_channel_amount` panics on `expect("channel info must exist in KVStore")`. The old file-based code excluded these because "pending" was in the file extension, which got stripped before the comparison.

src/ldk.rs
```diff
@@ -526,35 +544,62 @@ pub(crate) type OutputSweeper = ldk_sweep::OutputSweeper<
...
-            update_rgb_channel_amount(&channel_id_str, offered, received, ldk_data_dir, false);
-            break;
+    for inbound in [true, false] {
+        let namespace = if inbound {
+            RGB_PAYMENT_INFO_INBOUND_NS
+        } else {
+            RGB_PAYMENT_INFO_OUTBOUND_NS
+        };
+
+        if let Ok(keys) = kv_store.list(RGB_PRIMARY_NS, namespace) {
```
zoedberg: Using `list()` over both namespaces plus one `read()` per candidate, each running a `block_on` round trip seems inefficient. Since the function already gets the concrete `&SeaOrmKvStore`, you could do a single query returning key + value (something like `WHERE primary_namespace = 'rgb' AND secondary_namespace IN (...) AND key LIKE '%{payment_hash}'`)

src/utils.rs
```diff
@@ -349,6 +366,27 @@ pub(crate) async fn start_daemon(args: &UserArgs) -> Result<Arc<AppState>, AppEr
     let ldk_data_dir = args.storage_dir_path.join(LDK_DIR);
     let logger = Arc::new(FilesystemLogger::new(ldk_data_dir.clone()));
 
+    let db_path = get_db_path(&args.storage_dir_path);
```
zoedberg, line 407: `do_backup` zips the storage dir, which now includes the live DB while the connection pool is open, and writes can still happen while locked (`revoke_token` inserts from the auth middleware), so the backup can capture an inconsistent database.

IMO we can improve this by
- holding the state guard during the `/revoketoken` API to avoid concurrent runs with backup/restore
- run `VACUUM INTO '<staging>/rln_db'` on the live connection to get a transactionally consistent snapshot, zip that, and exclude the live `rln_db*` files from the directory walk.

Restore has the mirror problem: `restore_backup` unpacks over the DB file while the pool still holds the old one open, so existing connections keep reading the pre-restore inode. We should rebuild the connection pool after the restore.

src/disk.rs
```diff
@@ -73,76 +53,6 @@ impl Logger for FilesystemLogger {
...
-            Ok((pubkey, socket_addr)) => {
-                peer_data.insert(pubkey, socket_addr.expect("saved info with address"));
-            }
-            Err(e) => return Err(e),
-        }
-    }
-    Ok(peer_data)
-}
-
 pub(crate) fn read_network(
```
zoedberg: The background processor's persister is now the kv_store, so the network graph and scorer get persisted to sqlite, but startup still reads them from the old file paths via `disk::read_network` and `disk::read_scorer`. This means every restart silently discards the persisted graph and scorer. Please read them from the kv_store

src/utils.rs
```diff
@@ -349,6 +366,27 @@ pub(crate) async fn start_daemon(args: &UserArgs) -> Result<Arc<AppState>, AppEr
...
+    let connection_string = format!("sqlite:{}?mode=rwc", adjust_canonicalization(&db_path));
+    let mut opt = ConnectOptions::new(connection_string);
```
zoedberg: Could we set `journal_mode`, `synchronous` and `busy_timeout` explicitly?

src/ldk.rs
```diff
@@ -1814,13 +1883,14 @@ pub(crate) async fn start_ldk(
...
-    fs::write(
-        storage_dir_path.join(BITCOIN_NETWORK_FNAME),
-        bitcoin_network.to_string(),
-    )
-    .expect("able to write");
+    let kv_store = Arc::new(SeaOrmKvStore::from_connection(Arc::clone(
+        &static_state.database,
+    )));
+    kv_store.set_indexer_url(indexer_url)?;
+    kv_store.write_config(INDEXER_URL_FNAME, indexer_url);
```
zoedberg: Why are we duplicating config info into 2 different tables? I think we should keep a single source of truth for configuration params

zoedberg (changes requested):

src/utils.rs
dcorral: All done:
- /revoketoken now takes the state guard backup/restore hold, so no concurrent writes during either
- backup snapshots the DB with VACUUM INTO into staging and zips that; the live rln_db* files are excluded from the walk
- restore rebuilds the connection pool, re-runs migrations (so older backups get migrated), and reloads the in-memory revoked-token set from the restored DB

One detail: the pool must be closed before unpacking, not after, closing it afterwards lets sqlite's shutdown checkpoint overwrite the restored file through the still-open handle. 

Also fixed as part of this: the post-restore password check used to run against the pre-restore pool.

src/ldk.rs
dcorral: Agree, the kv copy only existed because LDK reads config through the KVStore interface; now SeaOrmKvStore maps reads of the wallet_config namespace directly onto the typed config table, nothing writes that namespace anymore, and write_config is gone from RgbKvStoreExt. 

Single source of truth is the config table; the LDK-side API is unchanged.

src/utils.rs
zoedberg: Perfect, thank you

src/ldk.rs
zoedberg: Great, thank you

zoedberg (changes requested):
It seems we are almost done, please rebase this on the updated `master` tip. If it's ok for you in the next iteration if I find something that needs to be addressed I will do it on top of your commit

PR #112, opened 2026-03-20
fix ChannelPending Panic

### LINKED ISSUE #111: rgb_send_lock gets stuck on channel open failure paths
`rgb-lightning-node` has multiple paths where `rgb_send_lock` gets stuck after channel open failures, blocking all subsequent `openchannel` calls with `OpenChannelInProgress`.

## Root cause

`rgb_send_lock` is set to `true` in the `openchannel` route but is not reliably released across all channel lifecycle event paths.

## Affected paths

1. **`Event::ChannelClosed`** — does not release the lock. When a channel is closed before funding (e.g. counterparty rejects during negotiation with `channel_funding_txo == None`), the lock stays stuck permanently.

2. **`Event::FundingGenerationReady`** — lock is only released inside the `funding_transaction_generated().is_err()` branch. When the call succeeds, the lock is held until `ChannelPending` fires.

3. **`Event::ChannelPending`** — spawns `rgb_send_btc_end`/`rgb_send_end` via `spawn_blocking` then `.await.unwrap()`. If the task panics, the lock stays stuck.

4. **No event fires** — when a peer disconnects before the `accept_channel` exchange completes, LDK may not emit any channel event, leaving the lock stuck permanently.

## Real-world trigger

Opening a channel to a peer that rejects it during negotiation (e.g. `maxValueInFlight too small`) causes `ChannelClosed` with no prior `FundingGenerationReady`. The lock is never released, and all subsequent `openchannel` calls return 403 `OpenChannelInProgress`.

## Fix

- **FundingGenerationReady**: release lock unconditionally after `funding_transaction_generated()` (both success and failure)
- **ChannelPending**: release lock before `spawn_blocking(send_btc_end)`; handle `JoinError` instead of `.unwrap()`
- **ChannelReady**: handle `JoinError` instead of `.unwrap()`
- **ChannelClosed**: release lock when `channel_funding_txo.is_none()`
- **Lock timeout**: change lock from `bool` to `Option<Instant>` with a 30-second timeout as safety net for cases where no LDK event fires

## Related

- RGB-Tools/rgb-lib#70 — `send_btc_end` panics on stale PSBT (caused by concurrent operations when lock is broken)

bitwalt:
reproduce and fix #111

### DIFF
--- a/src/test/channel_closed_lock_release.rs
+++ b/src/test/channel_closed_lock_release.rs
@@ -1,0 +1,116 @@
+use super::*;
+
+const TEST_DIR_BASE: &str = "tmp/channel_closed_lock_release/";
+
+/// Test that verifies the rgb_send_lock is released after a channel's peer
+/// disconnects before funding is generated.
+///
+/// Scenario: node1 opens a channel to node2, then immediately disconnects
+/// the peer. If the disconnect happens before FundingGenerationReady, LDK
+/// emits ChannelClosed with channel_funding_txo = None. Without the fix,
+/// the lock is never released and node1 cannot open new channels.
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn channel_closed_no_funding_releases_openchannel_lock() {
+    initialize();
+
+    let test_dir_node1 = format!("{TEST_DIR_BASE}node1");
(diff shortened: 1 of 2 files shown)


zoedberg:
@bitwalt I found a way to reproduce the bug in a deterministic way in https://github.com/RGB-Tools/rgb-lightning-node/pull/122/changes/57619a1316ff701e11d3f99a3589ad13c7b88aeb

PR #118, opened 2026-04-10
Add payment_hash support and improve HODL payment handling

free-free-6:
HODL Invoice Call Flow                                                                                                                            
                                                            
    C (Merchant) -- BTC Invoice --> A (Consumer/RGB User) -- swap request --> B (Gateway) -- BTC --> C
                                                                                      
  1. POST /lninvoice — Create HODL invoice with payment_hash (extracted from C's BTC invoice). HTLC will be held, not auto-claimed.                 
  2. A pays B's RGB invoice — PaymentClaimable event fires, payment stays Pending (no preimage).                                                    
  3. POST /sendpayment — B pays C's original BTC invoice. C claims it, exposing preimage.                                                           
  4. Auto-claim — PaymentSent event detects matching pending inbound payment, automatically calls claim_funds(preimage).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2239,6 +2239,12 @@ components:
             - integer
             - 'null'
           example: 42
+        payment_hash:
+          type:
+            - string
+            - 'null'
+          description: Optional payment hash (hex string, 32 bytes) for HODL invoices.
+          example: "229a76310b22a588271adcabc93362239c538103ebe2a8dea1822f9cde6c6040"
     LNInvoiceResponse:
       type: object
       required:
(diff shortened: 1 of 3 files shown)


zoedberg:
This doesn't seem to introduce/improve HODL invoices. HODL invoices are invoices that are not auto-claimed. This PR is just adding the possibility to decide the invoice payment hash, which makes sense only in the context of HODL invoices.
We already have a PR actually adding support to HODL invoices (https://github.com/RGB-Tools/rgb-lightning-node/pull/91), so this PR is not needed.

Issue #119, opened 2026-04-14
Swaps gets stuck on pending


Arshia-r-m:
I faced strange problem testing rapid swaps, after some points, swaps got stuck on `PENDING` state, the maker execute returned 200, but swaps never succeed/expired/failed.

From this I can assume the RLN never marks the swaps that are in pending state as expired even the expiry time has passed, I'll open a fix PR for this.

I also tried to find out why the swaps were stuck on pending and not failing/succeeding, but up to now I didn't find any thing, so can you please help me on this one? (probably could be an edge case)
I attached the directory of taker and maker both below, this happened on 2026-04-13 if you want to check.

Keep in mind that in between you are going to see some failed ones, pending stuck starts before, I don't think they are relevant.

Arshia-r-m:
I just check and swaps in pending will expire after 24h, but shouldn't they also expire after the expiry?
Let me know what you think

Arshia-r-m:
Since the file is over 25MB, I have uploaded it [here](https://drive.google.com/file/d/1avvrH6D8lmaTV6Egc4qgUPDA9ge97Lzt/view?usp=drive_link)

Arshia-r-m:
Update:
I probably have found the problem, the maker failed the payment with retry exhausted, and marked the payment as failed, but on the taker side, the `Event::HTLCHandlingFailed` has not been properly handled.

zoedberg:
When listing swaps, those in the waiting status move to an expired status if the expiration passed, those in the pending status move to a failed status if they are still pending after 24h after initiation time.
The `/makerexecute` API checks the expiration to be sure the expiration time set in the swapstring hasn't passed, if so it updates the status to expired.
The `/taker` API fails if the swapstring expiry time passed.
If the `/makerexecute` API succeeds the swap goes to a pending status. 
On the taker side the swap goes from the waiting status to a failed or succeeded status.
To me this seems the correct behavior, the expiration is a way for the maker to say "let's not initiate the swap after this time", so it's not an expiration tied up to the completion time but to the initiation time. IMO the parties should not interrupt the swap because it took time from when it started and when it completed (especially since, in case of swaps with multi-hop routes, collaboration across different parties is required and could take time).
Do you think swap expiration handling should change?

Would you please write a test that reproduces the problem you found?

Arshia-r-m:
What actually happened is that we have multiple swaps that are marked as failed on the maker side but pending on the taker side.

[...]

PR #121, opened 2026-04-17
add test for handling failed RGB forward swaps

Arshia-r-m:
prove the problem on #119


zoedberg:
Thank you for this! I also reproduced this bug with another test, with 2 swaps but not enough asset capacity to handle both. I have a fix that makes both tests pass but it's not very solid. I propose to leave this bug open for now since there are plans to replace the current swap logic with HODL invoices that would change some assumptions and it seems not worth it to spend time on this swap mechanism.

PR #124, opened 2026-04-29
Fix private-channel swap routing

### LINKED ISSUE #123: Private same-channel swaps fail with No route found
Swaps can fail when the maker needs to route a payment over a private channel that is only known locally.

In the same-channel swap flow, if the channel is opened as private, `/makerexecute` fails with:

```text
APIError: No route found
{"error":"No route found","code":403,"name":"NoRoute"}
```

## Why

The route lookup calls LDK route finding without passing local usable channels as `first_hops`. Public channels can be discovered through graph data, but private channels are not in the public graph. For a payment starting from our own node, LDK needs the local usable channels passed explicitly so it can route over private channels.

## Reproduction

Adjust the existing `swap_reverse_same_channel` test so it opens the same channel as private:

```rust
public: false
```

Then run:

```bash
cargo test test::swap_reverse_same_channel::swap_reverse_private_same_channel -- --exact --test-threads=1 --nocapture
```

Before the fix, the test fails at `/makerexecute` with `No route found`.

## Expected

Private same-channel swaps should execute successfully, same as public same-channel swaps.

## Proposed Fix

When computing a route from our own node, pass `channel_manager.list_usable_channels()` as `first_hops` to LDK's route finder. This allows local/private channels to be considered during route construction.

## Acceptance Criteria

- Add coverage in the existing same-channel swap test path.
- The test fails before the fix with `No route found`.
- The test passes after the fix.
- `cargo fmt --check` passes.
- `cargo check` passes.

bitwalt:
## Summary

- update the existing same-channel swap test to cover private-channel routing
- pass local usable channels as `first_hops` when route finding starts from our own node

## Root Cause

The swap route lookup did not provide LDK with local first-hop channels. That works for public channels discoverable from graph data, but private channels are not in the public graph, so route finding could fail with `No route found` even when a usable local private channel exists.

## Validation

- `cargo fmt --check`
- `cargo check`
- `cargo test test::swap_reverse_same_channel::swap_reverse_private_same_channel -- --exact --test-threads=1 --nocapture`

Fixes #123

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1222,6 +1222,7 @@ async fn open_channel_with_retry(
             None,
             None,
             true,
+            true,
         )
         .await;
 
@@ -1259,6 +1260,7 @@ async fn open_channel_funded_raw(
     fee_proportional_millionths: Option<u32>,
     temporary_channel_id: Option<&str>,
     with_anchors: bool,
+    public: bool,
 ) -> Result<Channel, Response> {
     open_channel_raw(
         node_address,
@@ -1273,6 +1275,7 @@ async fn open_channel_funded_raw(
         fee_proportional_millionths,
         temporary_channel_id,
         with_anchors,
+        public,
     )
     .await?;
 
@@ -1352,6 +1355,7 @@ async fn open_channel_raw(
     fee_proportional_millionths: Option<u32>,
     temporary_channel_id: Option<&str>,
     with_anchors: bool,
+    public: bool,
 ) -> Result<OpenChannelResponse, Response> {
     println!(
         "opening channel with {asset_amount:?} of asset {asset_id:?} from node {node_address} \
@@ -1383,7 +1387,7 @@ async fn open_channel_raw(
         asset_amount,
         asset_id: asset_id.map(|a| a.to_string()),
         push_asset_amount,
-        public: true,
+        public,
         with_anchors,
         fee_base_msat,
         fee_proportional_millionths,
@@ -1432,6 +1436,7 @@ async fn open_channel_with_custom_data(
         fee_proportional_millionths,
         temporary_channel_id,
         with_anchors,
+        true,
     )
     .await
     .expect("channel opening should succeed")
--- a/src/test/swap_reverse_same_channel.rs
+++ b/src/test/swap_reverse_same_channel.rs
@@ -186,4 +186,66 @@ async fn swap_reverse_same_channel() {
     assert_eq!(swaps_taker.taker.len(), 1);
     let swap_taker = swaps_taker.taker.first().unwrap();
     assert_eq!(swap_taker.status, SwapStatus::Succeeded);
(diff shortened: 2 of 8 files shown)


src/utils.rs
```diff
@@ -438,9 +444,22 @@ pub(crate) fn get_route(
...
         },
-        None,
+        first_hops.as_deref(),
         inflight_htlcs,
     );
 
     route.ok()
 }
+
+fn route_first_hops<'a>(
```
zoedberg: since this method is called only in one place I think we could just move its logic where it's used

src/utils.rs
```diff
@@ -415,6 +415,12 @@ pub(crate) fn get_route(
     hints: Vec<RouteHint>,
 ) -> Option<Route> {
     let inflight_htlcs = channel_manager.compute_inflight_htlcs();
+    let usable_channels = channel_manager.list_usable_channels();
+    let first_hops = route_first_hops(
+        &start,
+        channel_manager.get_our_node_id(),
+        usable_channels.iter(),
+    );
```
zoedberg:
```suggestion
    let usable_channels;
    let first_hops = if start == channel_manager.get_our_node_id() {
        usable_channels = channel_manager.list_usable_channels();
        Some(usable_channels.iter().collect::<Vec<_>>())
    } else {
        None
    };
```

src/test/mod.rs
```diff
@@ -1352,6 +1387,40 @@ async fn open_channel_raw(
...
+        fee_proportional_millionths,
+        temporary_channel_id,
+        with_anchors,
+        true,
+    )
+    .await
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn open_channel_raw_with_public(
```
zoedberg: no need to add this, let's add the `public` param to the `open_channel_raw` which is not used that much

src/test/swap_reverse_same_channel.rs
```diff
@@ -1,11 +1,11 @@
 use super::*;
 
-const TEST_DIR_BASE: &str = "tmp/swap_reverse_same_channel/";
+const TEST_DIR_BASE: &str = "tmp/swap_reverse_private_same_channel/";
 
 #[serial_test::serial]
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 #[traced_test]
-async fn swap_reverse_same_channel() {
+async fn swap_reverse_private_same_channel() {
```
zoedberg: please do not rename the test and directory and do not change the existing open channel. just add another private channel and swap at the end of this test

zoedberg (changes requested):
Thanks! Please address the requested changes and check the `swap_assets_liquidity_both_ways` test which seems to fail both on the CI and locally

bitwalt:
Changes should have been addressed

zoedberg (approved):
LGTM

Issue #125, opened 2026-05-01
Make bitcoind optional and support transaction-sync as an additional chain backend


bitwalt:
## Motivation

`rgb-lightning-node` currently requires a full `bitcoind` JSON-RPC connection in addition to an indexer (electrum/esplora). For most operators this is redundant: the indexer is already a complete chain source, and running an extra `bitcoind` purely to feed LDK doubles the storage footprint, the boot time, and the surface area for "what do I configure".

Removing it lets the node run against any maintained electrum/esplora endpoint (self-hosted, public, or an LSP-managed one), and shrinks the regtest/CI matrix accordingly. It also aligns the node with the deployment posture of most other RGB tooling, which is already indexer-only.

## Proposal

Drop the `bitcoind` requirement entirely and route block + transaction sync through the configured indexer:

- LDK chain source: replace `lightning-block-sync` (RPC mode) with `lightning-transaction-sync`, enabling both `electrum` and `esplora-blocking` features so a single binary supports both protocols.
- Fee estimation: poll the indexer (`estimate_fee` for electrum, `get_fee_estimates` for esplora), interpolating across confirmation buckets.
- Transaction broadcast: through the indexer's `transaction_broadcast` / esplora `POST /tx`.
- UTXO lookup for gossip verification: through the indexer (`blockchain.transaction.id_from_pos` / esplora block-index lookup), with async resolution so peer-manager isn't blocked.

## API / config surface changes

- **Removed** from `unlock` and `init` payloads (and the OpenAPI spec): `bitcoind_rpc_username`, `bitcoind_rpc_password`, `bitcoind_rpc_host`, `bitcoind_rpc_port`.
- **Unchanged**: `indexer_url` (already accepts `tcp://host:port` for electrum or `https://host` for esplora; auto-detected).
- **Dockerfile**: drops the `bitcoind` service from the runtime image; `compose.yaml` still ships `bitcoind` for regtest because we still need a block source for the test stack, just not for the node itself.
- **README**: regtest and docker examples updated to remove the four bitcoind RPC fields.

## Tradeoffs / behavioral notes

[...]

Issue #37, opened 2024-08-23
Maker execute fails with Error 403 (No Route Found) for High Amounts in Sats


bitwalt:
I've encountered a critical bug when attempting to perform a swap involving a higher amount in sats (e.g., 10,000 sats). The swap fails with the error `403 (No route found)`, even though there is sufficient liquidity (outbound_balance_msat and inbound_balance_msat) in the channel .

**Steps to Reproduce:**
1. Attempt a swap with a "high" amount in sats, such as 10,000 sats.
2. Observe that the `maker_execute` fails with the error `403 (No route found)` despite there being enough liquidity in the channel.

**Expected Behavior:**
The swap should succeed without any errors if there is enough liquidity in the channel.

**Actual Behavior:**
The swap fails with a `403 (No route found)` error.

**Additional Information:**
I tried replicating this issue by editing the `swap_roundtrip_buy_same_channel.rs` test. I set `qty_from` to 10,000,000 msat, and the test failed, confirming the issue.

zoedberg:
This is not a critical bug. It's more about lack of documentation and information returned by the APIs.

The `outbound_balance_msat` and `inbound_balance_msat` fields do not take into account the `htlc_maximum_msat`, which has an upper bound of 10% of the channel capacity (hard-coded in LDK). 
The only way to change this upper bound would be to manually change this value in our fork.
I'm not sure we should change that value though, since it's there as a safety measure to limit the node exposure to HTLCs.
But to improve the UX, we could expose the `next_outbound_htlc_limit_msat` and `next_outbound_htlc_minimum_msat` values in the list channels, so that the upper bound would be explicit and programs using RLN could deny users setting higher values when performing offchain transfers.

[...]

PR #126, opened 2026-05-01
Make bitcoind optional: add transaction-sync as an alternative chain backend

bitwalt:
Implements #125.

Reworked per @zoedberg's review: instead of replacing `lightning-block-sync` with `lightning-transaction-sync`, this **adds** transaction-sync as an alternative chain backend and keeps block-sync intact. The sync mode is chosen explicitly at unlock time.

### Sync modes

The `/unlock` payload gains a `sync_mode` tag:

- **`block_sync`** (default): consume full blocks from a trusted/local bitcoind over JSON-RPC (`lightning-block-sync`). The `bitcoind_rpc_*` parameters live inside this variant, so they are required exactly when this mode is selected and rejected otherwise — no manual "all or none" validation. This is the more trust-minimized / local-first option.
- **`transaction_sync`**: sync through the configured electrum/esplora `indexer_url` (`lightning-transaction-sync`). No bitcoind required — lighter to deploy for indexer-based setups.

### Cargo features

Each backend is gated behind a feature (`block-sync`, `transaction-sync`), **both enabled by default**, so a user who wants only one sync type can build with only the dependency they need. Both single-feature builds compile cleanly, and requesting a mode that wasn't compiled in returns a clear error.

### Implementation notes

- LDK is wired against trait objects for the fee estimator, broadcaster and gossip UTXO lookup, so a single set of type aliases serves both backends.
- Because `lightning-block-sync`'s `GossipVerifier` requires the `P2PGossipSync` to be typed with `Arc<Self>`, a small `BlockSyncGossipVerifier` provides the block-sync gossip UTXO lookup against the shared trait-object gossip sync (mirroring the transaction-sync `IndexerGossipVerifier`).

### Tests

Adds an integration test for the transaction-sync backend (as suggested): two nodes are unlocked in transaction-sync mode, an RGB asset channel is opened, a node is restarted (re-establishing the channel after syncing via the indexer) and the channel is cooperatively closed. The existing block-sync tests continue to pass unchanged.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -43,7 +43,13 @@ jobs:
         run: cargo build --release
       - name: Build (electrum only)
         if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features electrum
+        run: cargo build --release --no-default-features --features electrum,block-sync,transaction-sync
       - name: Build (esplora only)
         if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features esplora
+        run: cargo build --release --no-default-features --features esplora,block-sync,transaction-sync
+      - name: Build (block-sync only)
+        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
+        run: cargo build --release --no-default-features --features block-sync,electrum,esplora
+      - name: Build (transaction-sync only)
+        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
+        run: cargo build --release --no-default-features --features transaction-sync,electrum,esplora
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -41,15 +41,3 @@ jobs:
           rustflags: ""
       - name: Build
         run: cargo build --release
-      - name: Build (electrum only)
-        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features electrum,block-sync,transaction-sync
-      - name: Build (esplora only)
-        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features esplora,block-sync,transaction-sync
-      - name: Build (block-sync only)
-        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features block-sync,electrum,esplora
-      - name: Build (transaction-sync only)
-        if: matrix.name == 'linux' && matrix.toolchain == 'stable'
-        run: cargo build --release --no-default-features --features transaction-sync,electrum,esplora
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,11 +8,34 @@ rust-version = "1.88.0"
 name = "rgb-lightning-node"
 
 [features]
-default = ["electrum", "esplora"]
+default = [
+    "block-sync",
+    "transaction-sync",
+    "electrum",
+    "esplora",
+]
+# sync LDK from a bitcoind instance, consuming full blocks over JSON-RPC
+block-sync = [
+    "dep:lightning-block-sync",
+]
+# sync LDK from the indexer, without requiring a bitcoind instance
+transaction-sync = [
+    "dep:lightning-transaction-sync",
+]
 # support indexers implementing the electrum protocol
-electrum = ["rgb-lib/electrum", "lightning/electrum"]
+electrum = [
+    "rgb-lib/electrum",
+    "lightning/electrum",
+    "lightning-transaction-sync?/electrum",
+    "dep:electrum-client",
+]
 # support indexers implementing the esplora protocol
-esplora = ["rgb-lib/esplora", "lightning/esplora"]
+esplora = [
+    "rgb-lib/esplora",
+    "lightning/esplora",
+    "lightning-transaction-sync?/esplora-blocking",
+    "dep:esplora-client",
+]
 
 [dependencies]
 amplify = { version = "=4.8.1", default-features = false }
@@ -29,17 +52,20 @@ chacha20poly1305 = { version = "0.10.1", features = ["stream"] }
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 clap = "4.5.20"
 dirs = "5.0.1"
+electrum-client = { version = "0.24.0", default-features = false, features = ["use-rustls"], optional = true }
+esplora-client = { version = "0.12", default-features = false, features = ["blocking-https-rustls"], optional = true }
 futures = "0.3"
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false }
 lightning = { version = "0.2.0", path = "./rust-lightning/lightning", features = ["dnssec"] }
 lightning-background-processor = { version = "0.2.0", path = "./rust-lightning/lightning-background-processor" }
-lightning-block-sync = { version = "0.2.0", features = ["rpc-client", "tokio"] }
+lightning-block-sync = { version = "0.2.0", features = ["rpc-client", "tokio"], optional = true }
 lightning-dns-resolver = { version = "0.3.0", path = "./rust-lightning/lightning-dns-resolver" }
 lightning-invoice = { version = "0.34.0", features = ["std"], path = "./rust-lightning/lightning-invoice" }
 lightning-macros = { version = "0.2.0" }
 lightning-net-tokio = { version = "0.2.0" }
 lightning-persister = { version = "0.2.0", path = "./rust-lightning/lightning-persister", features = ["tokio"] }
 lightning-rapid-gossip-sync = { version = "0.2.0", path = "./rust-lightning/lightning-rapid-gossip-sync" }
+lightning-transaction-sync = { version = "0.2.0", path = "./rust-lightning/lightning-transaction-sync", optional = true }
 rand = "0.8.5"
 regex = { version = "1.11", default-features = false }
 rgb-lib = { version = "=0.3.0-beta.7", default-features = false }
@@ -61,7 +87,7 @@ zip = { version = "2.2.0", default-features = false, features = ["time", "zstd"]
 
 [dev-dependencies]
 dircmp = "0.2.0"
-electrum-client = "0.20.0"
+electrum-client = "0.24.0"
 http = "1.4.0"
 lazy_static = { version = "1.5.0", default-features = false }
 lightning = { version = "0.2.0", path = "./rust-lightning/lightning", features = ["_rln_test_hooks"] }
--- a/README.md
+++ b/README.md
@@ -160,9 +160,7 @@ For more info about regtest utility commands, run:
 ./regtest.sh -h
 ```
 
-See [Sync modes](#sync-modes) for how the `/unlock` payload selects the chain
-backend. When unlocking regtest nodes with `BlockSync` use the following local
-services:
+When unlocking regtest nodes use the following local services:
 - bitcoind_rpc_username: user
 - bitcoind_rpc_password: password
 - bitcoind_rpc_host: localhost
@@ -175,9 +173,6 @@ To unlock a regtest nodes running in docker use the following local services:
 - bitcoind_rpc_host: bitcoind
 - bitcoind_rpc_port: 18443
 - indexer_url: electrs:50001
-
-To sync through the indexer instead, select the `TransactionSync` mode and omit
-the `bitcoind_rpc_*` parameters.
 
 ### Testnet
 
@@ -433,8 +428,8 @@ time via the `ldk_chain_sync` field of the `/unlock` payload (see the
   is the more trust-minimized option, since the node does not rely on an indexer
   to tell it which transactions are relevant.
 - `TransactionSync`: sync through an electrum/esplora indexer, so no `bitcoind`
-  is needed. By default this reuses the wallet's `indexer_url`; a dedicated
-  indexer for LDK can be set under this mode's `config` via `indexer_url`.
+  is needed. The indexer LDK syncs against is given under this mode's `config`
+  via `indexer_url` and can differ from the one the RGB wallet uses.
 
 Both modes are available in a stock build. See
 [Chain sync support](#chain-sync-support) to build with only one of them.
--- a/README.md
+++ b/README.md
@@ -50,18 +50,34 @@ Support for the indexer protocols is behind cargo features, `electrum` and
 
 To support electrum indexers only:
 ```sh
-cargo install --locked --path . --no-default-features --features electrum
+cargo install --locked --path . --no-default-features --features electrum,block-sync,transaction-sync
 ```
 
 To support esplora indexers only:
 ```sh
-cargo install --locked --path . --no-default-features --features esplora
+cargo install --locked --path . --no-default-features --features esplora,block-sync,transaction-sync
+```
+
+### Chain sync support
+
+Support for the chain sync backends is behind cargo features, `block-sync` and
+`transaction-sync` (both enabled by default). At least one of them needs to be
+enabled. See [Sync modes](#sync-modes) for what each backend does.
+
+To support the block-sync backend only:
+```sh
+cargo install --locked --path . --no-default-features --features block-sync,electrum,esplora
+```
+
+To support the transaction-sync backend only:
+```sh
+cargo install --locked --path . --no-default-features --features transaction-sync,electrum,esplora
 ```
 
 ## Run
 
 In order to operate, the node will need:
-- a bitcoind node
+- a bitcoind node (only for the `BlockSync` [sync mode](#sync-modes))
 - an indexer instance (electrum or esplora)
 
 Once services are running, daemons can be started.
@@ -144,7 +160,9 @@ For more info about regtest utility commands, run:
 ./regtest.sh -h
 ```
 
-When unlocking regtest nodes use the following local services:
+See [Sync modes](#sync-modes) for how the `/unlock` payload selects the chain
+backend. When unlocking regtest nodes with `BlockSync` use the following local
+services:
 - bitcoind_rpc_username: user
 - bitcoind_rpc_password: password
 - bitcoind_rpc_host: localhost
@@ -157,6 +175,9 @@ To unlock a regtest nodes running in docker use the following local services:
 - bitcoind_rpc_host: bitcoind
 - bitcoind_rpc_port: 18443
 - indexer_url: electrs:50001
+
+To sync through the indexer instead, select the `TransactionSync` mode and omit
+the `bitcoind_rpc_*` parameters.
 
 ### Testnet
 
@@ -400,6 +421,23 @@ Example proxy URLs (only when using proxy transport):
 |-------------|-----------------------|
 | Local | `rpc://127.0.0.1:3000/json-rpc` |
 | Public | `rpcs://proxy.iriswallet.com/0.2/json-rpc` |
+
+## Sync modes
+
+The node keeps LDK in sync with the chain in one of two ways, selected at unlock
+time via the `ldk_chain_sync` field of the `/unlock` payload (see the
+`UnlockRequest` schema in `openapi.yaml` for the exact shape):
+
+- `BlockSync`: consume full blocks from a trusted/local `bitcoind` over JSON-RPC.
+  The `bitcoind_rpc_*` parameters are provided under this mode's `config`. This
+  is the more trust-minimized option, since the node does not rely on an indexer
+  to tell it which transactions are relevant.
+- `TransactionSync`: sync through an electrum/esplora indexer, so no `bitcoind`
+  is needed. By default this reuses the wallet's `indexer_url`; a dedicated
+  indexer for LDK can be set under this mode's `config` via `indexer_url`.
+
+Both modes are available in a stock build. See
+[Chain sync support](#chain-sync-support) to build with only one of them.
 
 ## Test
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2277,11 +2277,11 @@ components:
           enum: [TransactionSync]
         config:
           type: object
+          required:
+            - indexer_url
           properties:
             indexer_url:
-              type:
-                - string
-                - 'null'
+              type: string
               example: 127.0.0.1:50001
     ListAssetsRequest:
       type: object
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2228,6 +2228,61 @@ components:
           example: 89d28bd306aa9bb906fd0ac31092d04c37c919a171b343083167e2a3cdc60578
         status:
           $ref: '#/components/schemas/HTLCStatus'
+    LdkChainSync:
+      oneOf:
+        - $ref: '#/components/schemas/LdkChainSyncBlockSync'
+        - $ref: '#/components/schemas/LdkChainSyncTransactionSync'
+      discriminator:
+        propertyName: mode
+        mapping:
+          BlockSync: '#/components/schemas/LdkChainSyncBlockSync'
+          TransactionSync: '#/components/schemas/LdkChainSyncTransactionSync'
+    LdkChainSyncBlockSync:
+      type: object
+      required:
+        - mode
+        - config
+      properties:
+        mode:
+          type: string
+          enum: [BlockSync]
+        config:
+          type: object
+          required:
+            - bitcoind_rpc_username
+            - bitcoind_rpc_password
+            - bitcoind_rpc_host
+            - bitcoind_rpc_port
+          properties:
+            bitcoind_rpc_username:
+              type: string
+              example: user
+            bitcoind_rpc_password:
+              type: string
+              example: password
+            bitcoind_rpc_host:
+              type: string
+              example: localhost
+            bitcoind_rpc_port:
+              type: integer
+              example: 18443
+    LdkChainSyncTransactionSync:
+      type: object
+      required:
+        - mode
+        - config
+      properties:
+        mode:
+          type: string
+          enum: [TransactionSync]
+        config:
+          type: object
+          properties:
+            indexer_url:
+              type:
+                - string
+                - 'null'
+              example: 127.0.0.1:50001
     ListAssetsRequest:
       type: object
       required:
@@ -3436,28 +3491,15 @@ components:
       type: object
       required:
         - password
-        - bitcoind_rpc_username
-        - bitcoind_rpc_password
-        - bitcoind_rpc_host
-        - bitcoind_rpc_port
+        - ldk_chain_sync
         - indexer_url
         - announce_addresses
       properties:
         password:
           type: string
           example: nodepassword
-        bitcoind_rpc_username:
-          type: string
-          example: user
-        bitcoind_rpc_password:
-          type: string
-          example: password
-        bitcoind_rpc_host:
-          type: string
-          example: localhost
-        bitcoind_rpc_port:
-          type: integer
-          example: 18443
+        ldk_chain_sync:
+          $ref: '#/components/schemas/LdkChainSync'
         indexer_url:
           type: string
           example: 127.0.0.1:50001
--- a/src/error.rs
+++ b/src/error.rs
@@ -4,7 +4,9 @@ use axum::{
     response::{IntoResponse, Response},
     Json,
 };
-use rgb_lib::{BitcoinNetwork, Error as RgbLibError};
+#[cfg(feature = "block-sync")]
+use rgb_lib::BitcoinNetwork;
+use rgb_lib::Error as RgbLibError;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -74,6 +76,7 @@ pub enum APIError {
     #[error("Failed to sync BDK: {0}")]
     FailedBdkSync(String),
 
+    #[cfg(feature = "block-sync")]
     #[error("Failed to connect to bitcoind client: {0}")]
     FailedBitcoindConnection(String),
 
@@ -263,6 +266,7 @@ pub enum APIError {
     #[error("Network error: {0}")]
     Network(String),
 
+    #[cfg(feature = "block-sync")]
     #[error("The network of the given bitcoind ({0}) doesn't match the node's chain ({1})")]
     NetworkMismatch(String, BitcoinNetwork),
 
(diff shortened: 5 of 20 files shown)


zoedberg (changes requested):
Please check my comment https://github.com/RGB-Tools/rgb-lightning-node/issues/125#issuecomment-4380642107, I explained why we should not replace the block-sync with the transaction-sync but add it instead as an alternative sync mode. Please update this PR accordingly

zoedberg:
Hi @bitwalt, here https://github.com/RGB-Tools/rgb-lightning-node/issues/125#issuecomment-4381346472 you said you agreed in updating this PR to add this as an alternative sync mode. Are you still willing to complete this PR?

bitwalt:
Hi Zoe, sure I'll try to complete this

bitwalt:
Thanks @zoedberg — I've reworked the PR along the lines you suggested.

Rather than replacing block-sync, it now keeps `lightning-block-sync` and **adds** `lightning-transaction-sync` as an alternative, selectable via an explicit `sync_mode` on the unlock request:

- `block_sync` (default) — full blocks from a trusted/local bitcoind. The `bitcoind_rpc_*` params are embedded in this variant so they're required exactly when this mode is chosen (no manual all-or-none checking).
- `transaction_sync` — sync through the configured electrum/esplora indexer only, no bitcoind.

Both are behind Cargo features (`block-sync`, `transaction-sync`), enabled by default, so a user can compile just the sync dependency they want; both single-feature builds compile cleanly.

I added the integration test you described (unlock 2 nodes, open a channel, restart a node, close the channel) exercising the transaction-sync path, and the existing block-sync tests still pass.

Two notes for review:
1. The `/unlock` payload shape changes (the bitcoind params now sit under the `sync_mode: block_sync` variant). You mentioned not worrying about existing setups in this phase, but happy to reshape if you'd prefer a different layout.
2. For the block-sync gossip UTXO lookup I added a small `BlockSyncGossipVerifier` instead of `lightning-block-sync`'s `GossipVerifier`, because the latter requires the `P2PGossipSync` to be typed with `Arc<Self>`, which is incompatible with the trait-object gossip sync needed to serve both backends from one `PeerManager` type. Let me know if you'd rather approach that differently.

README.md
```diff
@@ -137,13 +152,17 @@ When unlocking regtest nodes use the following local services:
...
+- sync_mode: block_sync
 - bitcoind_rpc_username: user
 - bitcoind_rpc_password: password
 - bitcoind_rpc_host: bitcoind
 - bitcoind_rpc_port: 18443
 - indexer_url: electrs:50001
 - proxy_endpoint: rpc://proxy:3000/json-rpc
 
+To sync through the indexer instead, set `sync_mode: transaction_sync` and omit
+the `bitcoind_rpc_*` parameters (`indexer_url` is still required).
```
zoedberg:
```suggestion
To sync through the indexer instead, set `sync_mode: transaction_sync` and omit
the `bitcoind_rpc_*` parameters.
```
since `transaction_sync` uses the indexer I think it's confusing to say the indexer url is still required

openapi.yaml
```diff
@@ -3252,26 +3252,39 @@ components:
...
-        - bitcoind_rpc_port
+        - sync_mode
         - announce_addresses
       properties:
         password:
           type: string
           example: nodepassword
+        sync_mode:
+          type: string
+          description: >
```
zoedberg: Please avoid `description`s of single fields, they deviate from the style we used so far. We'll probably add description to all object fields in a single future commit

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+/// Selects how LDK keeps in sync with the chain.
+///
+/// The bitcoind RPC parameters live inside the `block_sync` variant so that
+/// they are required exactly when that mode is selected and rejected
+/// otherwise, without any manual "all or none" validation. The mode is chosen
+/// explicitly via the `sync_mode` tag rather than being inferred from which
+/// fields are populated.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(tag = "sync_mode", rename_all = "snake_case")]
+pub(crate) enum ChainSource {
```
zoedberg: please keep the alphabetical order

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+}
+
 #[derive(Deserialize, Serialize)]
 pub(crate) struct UnlockRequest {
     pub(crate) password: String,
-    pub(crate) bitcoind_rpc_username: String,
-    pub(crate) bitcoind_rpc_password: String,
-    pub(crate) bitcoind_rpc_host: String,
-    pub(crate) bitcoind_rpc_port: u16,
+    #[serde(flatten)]
```
```suggestion
```
Please also fix openapi accordingly

README.md
```diff
@@ -137,13 +152,17 @@ When unlocking regtest nodes use the following local services:
 - proxy_endpoint: rpc://127.0.0.1:3000/json-rpc
 
 To unlock a regtest nodes running in docker use the following local services:
+- sync_mode: block_sync
```
zoedberg: This parameter is for all networks, not only for regtest, therefore I would not add this field here, this section is dedicated to the network-specific parameters

README.md
```diff
@@ -128,7 +128,22 @@ For more info about regtest utility commands, run:
...
+  `bitcoind` over JSON-RPC. Requires the `bitcoind_rpc_*` parameters. This is
+  the more trust-minimized option since the node does not rely on the indexer
+  to tell it which transactions are relevant.
+- `transaction_sync`: sync through the configured `indexer_url`
+  (electrum/esplora) only. No `bitcoind_rpc_*` parameters are needed. This is
+  lighter to deploy and fits indexer-based setups.
+
+Each sync mode is gated behind a Cargo feature (`block-sync` and
+`transaction-sync`); both are enabled by default, so a stock build supports
+either mode at unlock time.
```
zoedberg: This is not regtest-specific, we should instead add a dedicated section to explain the library features

README.md
```diff
@@ -128,7 +128,22 @@ For more info about regtest utility commands, run:
 ./regtest.sh -h
 ```
 
-When unlocking regtest nodes use the following local services:
+The `/unlock` payload selects how the node syncs with the chain via the
+`sync_mode` field:
+- `block_sync` (default build): consume full blocks from a trusted/local
```
zoedberg: I see both features are set in the library default features set, which is correct, so I think saying block_sync is the default build is incorrect

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+    BlockSync {
+        bitcoind_rpc_username: String,
+        bitcoind_rpc_password: String,
+        bitcoind_rpc_host: String,
+        bitcoind_rpc_port: u16,
+    },
+    /// Sync through the configured electrum/esplora indexer
+    /// (`lightning-transaction-sync`). Requires the `transaction-sync` build
+    /// feature. Uses `indexer_url` as the chain source, so no bitcoind is
+    /// needed.
```
```suggestion
```
let's avoid docs on private objects

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+/// The bitcoind RPC parameters live inside the `block_sync` variant so that
+/// they are required exactly when that mode is selected and rejected
+/// otherwise, without any manual "all or none" validation. The mode is chosen
+/// explicitly via the `sync_mode` tag rather than being inferred from which
+/// fields are populated.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(tag = "sync_mode", rename_all = "snake_case")]
+pub(crate) enum ChainSource {
+    /// Consume full blocks from a trusted/local bitcoind over JSON-RPC
+    /// (`lightning-block-sync`). Requires the `block-sync` build feature.
```
```suggestion
```
let's avoid docs on private objects

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+    BlockSync {
```
zoedberg: Please feature gate the variants, it would simplify the code, it's acceptable to lose `APIError::UnsupportedSyncMode`

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
     JsonRpc,
 }
 
+/// Selects how LDK keeps in sync with the chain.
+///
+/// The bitcoind RPC parameters live inside the `block_sync` variant so that
+/// they are required exactly when that mode is selected and rejected
+/// otherwise, without any manual "all or none" validation. The mode is chosen
+/// explicitly via the `sync_mode` tag rather than being inferred from which
+/// fields are populated.
```
```suggestion
```
let's avoid docs on private objects

src/routes.rs
```suggestion
#[serde(tag = "mode", content = "config")]
pub(crate) enum LdkChainSync {
```
this maps good into openapi and IMO is clearer

src/routes.rs
```diff
@@ -1359,13 +1359,36 @@ pub(crate) enum TransportType {
...
+        bitcoind_rpc_username: String,
+        bitcoind_rpc_password: String,
+        bitcoind_rpc_host: String,
+        bitcoind_rpc_port: u16,
+    },
+    /// Sync through the configured electrum/esplora indexer
+    /// (`lightning-transaction-sync`). Requires the `transaction-sync` build
+    /// feature. Uses `indexer_url` as the chain source, so no bitcoind is
+    /// needed.
+    TransactionSync,
```
zoedberg: I would add the indexer to this variant, so that a user could use two different indexer sources for LN and for the RGB wallet

src/indexer.rs
```diff
@@ -0,0 +1,687 @@
...
+fn sync_electrum_registered_txs(
+    client: Arc<ElectrumClient>,
+    registered_txs: &Mutex<HashMap<Txid, RegisteredTx>>,
+    confirmables: &[Confirmable],
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let mut confirmed = Vec::new();
+    let mut unconfirmed = Vec::new();
+
+    {
+        let mut registered_txs = registered_txs.lock().unwrap();
```
zoedberg: I don't think we should keep the mutex locked while doing blocking network calls

src/ldk_chain_backend/block_sync.rs
```diff
@@ -356,3 +361,103 @@ impl BroadcasterInterface for BitcoindClient {
...
+/// UTXO lookup for gossip verification backed by bitcoind.
+///
+/// This is the block-sync counterpart to the transaction-sync
+/// `IndexerGossipVerifier`. Unlike `lightning-block-sync`'s own
+/// `GossipVerifier` (which requires the `P2PGossipSync` to be typed with
+/// `Arc<Self>` as its UTXO lookup), this implementation resolves against a
+/// `PeerGossipSync` whose lookup is a trait object, so a single `PeerManager`
+/// type can serve both sync backends. The funding output for a channel
+/// announcement is resolved from full blocks fetched from bitcoind.
+pub(crate) struct BlockSyncGossipVerifier {
```
zoedberg, line 312: since you had to reimplement it, I would port its `block_cache` too

src/ldk_chain_backend/transaction_sync.rs
```diff
@@ -0,0 +1,687 @@
...
+
+    fn register_output(&self, output: WatchedOutput) {
+        match self {
+            Self::Electrum { client, .. } => client.register_output(output),
+            Self::Esplora(client) => client.register_output(output),
+        }
+    }
+}
+
+fn sync_electrum_registered_txs(
```
zoedberg: this is duplicating work that upstream `ElectrumSyncClient::sync` already performs, so the parallel `registered_txs` map and the second pass appear redundant. either delete it and rely on upstream or, if it's compensating for a concrete failure case, put that reason in a comment

src/indexer.rs
```diff
@@ -0,0 +1,687 @@
...
+                })?;
+                let tip_height = client.get_height().map_err(|e| {
+                    io::Error::other(format!("failed to fetch esplora tip height: {e}"))
+                })?;
+                Ok(BestBlock::new(tip_hash, tip_height))
+            }
+        }
+    }
+
+    fn lookup_utxo(
```
zoedberg: this is skipping the confirmation check that the block-sync path enforces (< 6 confirmations). this makes the two backends behave differently for gossip verification, IMO worth aligning

src/ldk_chain_backend/transaction_sync.rs
```diff
@@ -0,0 +1,687 @@
...
+                Ok(BestBlock::new(tip.header.block_hash(), tip.height as u32))
+            }
+            IndexerBackend::Esplora(client) => {
+                let tip_hash = client.get_tip_hash().map_err(|e| {
+                    io::Error::other(format!("failed to fetch esplora tip hash: {e}"))
+                })?;
+                let tip_height = client.get_height().map_err(|e| {
+                    io::Error::other(format!("failed to fetch esplora tip height: {e}"))
+                })?;
+                Ok(BestBlock::new(tip_hash, tip_height))
```
zoedberg, line 144: this is not atomic, a block that arrives between the two calls yields to a hash/height mismatch in the `BestBlock`

Cargo.toml
```diff
@@ -7,6 +7,19 @@ rust-version = "1.88.0"
 [[bin]]
 name = "rgb-lightning-node"
 
+[features]
```
zoedberg, line 10: we should make sure build with single-features works in `.github/workflows/build.yaml`. and since one of the features is always needed we should add the following to `src/main.rs`:
```
#[cfg(not(any(
    feature = "block-sync,
    feature = "transaction-sync"
)))]
compile_error!("You must enable at least 'block-sync' or 'transaction-sync'");
```

src/test/transaction_sync.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Exercises the transaction-sync chain backend end-to-end: two nodes are
+/// unlocked in transaction-sync mode (syncing through the indexer, with no
+/// bitcoind RPC access), an RGB asset channel is opened between them, a node is
+/// restarted (so it must re-establish the channel after syncing via the
+/// indexer) and finally the channel is cooperatively closed.
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn transaction_sync() {
```
zoedberg: the esplora backend is entirely untested, please add an esplora service to the compose file and test the transaction sync also with that indexer

zoedberg (changes requested):
Thanks for updating this PR! Please check the requested changes and sign the commit

bitwalt:
Thanks for the thorough review @zoedberg — reworked and squashed into a single **signed** commit. Rundown per thread:

### `chain_source` (was `ChainSource`) — `routes.rs` / `openapi.yaml`
- Renamed to `LdkChainSync` and switched to `#[serde(tag = "mode", content = "config")]`. I kept `rename_all = "snake_case"` so the wire values stay `block_sync` / `transaction_sync` — dropping it flips them to `BlockSync` / `TransactionSync`, which would diverge from the README/openapi. Happy to drop it if you'd rather match the suggestion literally.
- Removed the doc comments on the (private) enum and its variants.
- Feature-gated both variants and dropped `APIError::UnsupportedSyncMode` accordingly.
- Removed `#[serde(flatten)]` — `chain_source` is now a nested object; `openapi.yaml` updated to match and all the per-field `description`s I'd added are gone.
- Moved the type to its alphabetical position.
- Added an optional `indexer_url` to the `transaction_sync` `config`, so the RGB wallet and LDK can use different indexer sources.

### `indexer.rs`
- esplora `get_best_block` is now atomic: the height is derived from the tip hash we just fetched (`get_block_status`), so there's no hash/height mismatch if a block arrives mid-call.
- The transaction-sync gossip lookup now enforces the same six-confirmation depth as the block-sync path.
- On the "redundant electrum pass" (`sync_electrum_registered_txs`): I first deleted it and relied on `ElectrumSyncClient::sync`, but that regresses — the electrum `transaction_sync` test then deterministically hangs at funding lock-in (the channel never reaches `channel_ready`), while esplora is fine. So it's compensating for a concrete failure rather than being redundant. I kept it with a comment explaining exactly that, and reworked it so the `registered_txs` lock is **not** held across the electrum network calls (your other comment).

### `bitcoind.rs`
- Ported upstream `GossipVerifier`'s block cache into `BlockSyncGossipVerifier`.

### Build / features
- Added the `compile_error!` guard in `main.rs` when neither feature is enabled, and a dedicated `build-features` CI job that builds each single feature.

### `README.md`
- Moved the sync-mode + Cargo-feature explanation into a dedicated **Sync modes** section, dropped the "(default build)" wording, removed the `sync_mode` bullets from the network-specific service lists, and fixed the indexer wording.

### esplora coverage
- The esplora backend was untested — added a `blockstream/esplora` service to `compose.yaml` (+ `regtest.sh` wiring) and split the test into `transaction_sync_electrum` and `transaction_sync_esplora`. The esplora variant also exercises the dedicated-LN-indexer path (LDK on esplora, wallet on electrum).

### Verification
- Builds with default / `block-sync`-only / `transaction-sync`-only, and correctly fails with neither feature.
- `cargo clippy -D warnings` and `cargo fmt --check` clean across those feature sets.
- Both integration tests (`transaction_sync_electrum`, `transaction_sync_esplora`) pass against a real regtest bitcoind + electrs + esplora.

.github/workflows/build.yaml
```diff
@@ -36,3 +36,26 @@ jobs:
           rustflags: ""
       - name: Build
         run: cargo build --release
+
+  build-features:
+    # Each sync backend is gated behind a Cargo feature; make sure the crate
+    # still builds with only one of them enabled.
```
zoedberg:
```suggestion
  build-features:
```

src/test/mod.rs
```diff
@@ -270,6 +274,21 @@ async fn start_node(
...
+        block_sync_source(),
+    )
+    .await
+}
+
+async fn start_node_with_sync(
+    node_test_dir: &str,
+    node_peer_port: u16,
+    keep_node_dir: bool,
+    chain_source: LdkChainSync,
```
zoedberg:
```suggestion
    ldk_chain_sync: LdkChainSync,
```
Let's avoid two names for the same concept, here and in all other places where you are still using `chain_source`

src/test/transaction_sync.rs
```diff
@@ -0,0 +1,101 @@
...
+async fn transaction_sync_electrum() {
+    transaction_sync_roundtrip(TEST_DIR_BASE, None).await;
+}
+
+/// Transaction-sync against the esplora indexer, pointing LDK at a dedicated
+/// esplora source while the RGB wallet keeps using electrum.
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn transaction_sync_esplora() {
```
zoedberg, line 152: Please gate the tests with the `transaction-sync` feature

src/error.rs
```diff
@@ -56,6 +56,8 @@ pub enum APIError {
     #[error("Failed to sync BDK: {0}")]
     FailedBdkSync(String),
 
+    // Only constructed by the block-sync backend.
+    #[cfg_attr(not(feature = "block-sync"), allow(dead_code))]
```
zoedberg: Please do not allow `dead_code` (same for `NetworkMismatch`)
```suggestion
    #[cfg(feature = "block-sync")]
```

src/ldk_chain_backend/transaction_sync.rs
```diff
@@ -0,0 +1,752 @@
...
+                })?;
+                Ok(BestBlock::new(tip.header.block_hash(), tip.height as u32))
+            }
+            IndexerBackend::Esplora(client) => {
+                let tip_hash = client.get_tip_hash().map_err(|e| {
+                    io::Error::other(format!("failed to fetch esplora tip hash: {e}"))
+                })?;
+                // Derive the height from the tip hash we just fetched (rather
+                // than a separate `get_height` call) so the hash and height stay
+                // consistent even if a new block arrives between requests.
```
zoedberg:
```suggestion
```

openapi.yaml
```diff
@@ -3252,27 +3252,44 @@ components:
...
-          example: password
-        bitcoind_rpc_host:
-          type: string
-          example: localhost
-        bitcoind_rpc_port:
-          type: integer
-          example: 18443
+        chain_source:
+          type: object
+          required:
```
zoedberg: This is incorrect. You should use `discriminator` with `mapping`, see `Assignment` for an example on how to use it

src/indexer.rs
```diff
@@ -0,0 +1,752 @@
...
+        &self,
+        chain_hash: ChainHash,
+        short_channel_id: u64,
+    ) -> Result<TxOut, UtxoLookupError> {
+        if chain_hash != ChainHash::using_genesis_block(self.network) {
+            return Err(UtxoLookupError::UnknownChain);
+        }
+
```
zoedberg:
```suggestion
```

I think we should keep it symmetric to the block-sync lookup, there the chain hash gets ignored because in `rust-lightning/lightning/src/routing/gossip.rs#L2197` LDK already validates the chain hash correctness

src/indexer.rs
```diff
@@ -0,0 +1,752 @@
...
+            None => Err(UtxoLookupError::UnknownTx),
+        }
+    }
+}
+
+impl UtxoLookup for IndexerClient {
+    fn get_utxo(&self, chain_hash: &ChainHash, short_channel_id: u64) -> UtxoResult {
+        UtxoResult::Sync(self.lookup_utxo(*chain_hash, short_channel_id))
+    }
+}
```
zoedberg: This seems unused, if so please remove it and remove the `lookup_utxo` method (move logic to `IndexerGossipVerifier::get_utxo`)

src/routes.rs
```diff
@@ -766,6 +766,20 @@ pub(crate) struct KeysendResponse {
     pub(crate) status: HTLCStatus,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(tag = "mode", content = "config", rename_all = "snake_case")]
```
```suggestion
#[serde(tag = "mode", content = "config")]
```
Using snake case here deviates from the style we used so far (e.g. `JsonRpc`, `Nia`)

src/test/transaction_sync.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Exercises the transaction-sync chain backend end-to-end: two nodes are
+/// unlocked in transaction-sync mode (syncing through the indexer, with no
+/// bitcoind RPC access), an RGB asset channel is opened between them, a node is
+/// restarted (so it must re-establish the channel after syncing via the
+/// indexer) and finally the channel is cooperatively closed.
+///
+/// `ln_indexer_url` selects the indexer LDK syncs against: `None` reuses the
+/// wallet's `indexer_url` (electrum), while `Some(..)` points LDK at a
+/// dedicated indexer (used to cover the esplora backend).
+async fn transaction_sync_roundtrip(test_dir_base: &str, ln_indexer_url: Option<String>) {
```
zoedberg: Please extend the test. To cover the gossip lookup we need a 3-node multihop RGB payment.

README.md
```diff
@@ -43,10 +43,54 @@ The docker image can be built with:
 docker build -t rgb-lightning-node .
 ```
 
+## Sync modes
```
zoedberg: I think this should move down, in between `Use` and `Test`

README.md
```diff
@@ -43,10 +43,54 @@ The docker image can be built with:
...
+  "chain_source": { "mode": "transaction_sync", "config": {} }
+}
+```
+
+Each sync mode is gated behind a Cargo feature (`block-sync` and
+`transaction-sync`), both enabled by default, so a stock build supports either
+mode at unlock time. To build with a single backend (and pull in only its
+dependencies), disable the defaults and enable the one you need, e.g.:
+```sh
+cargo build --release --no-default-features --features transaction-sync
```
zoedberg: Since this is a binary we should use install here. Also add a reference to this section somewhere in the `Install` section

README.md
```diff
@@ -43,10 +43,54 @@ The docker image can be built with:
...
+}
+```
+or to sync through the indexer:
+```json
+{
+  "password": "...",
+  "chain_source": { "mode": "transaction_sync", "config": {} }
+}
+```
+
```
zoedberg: For examples we already have openapi so let's drop this

src/indexer.rs
```diff
@@ -0,0 +1,752 @@
...
+    );
+    fees.insert(
+        ConfirmationTarget::ChannelCloseMinimum,
+        AtomicU32::new(MIN_FEERATE),
+    );
+    fees.insert(
+        ConfirmationTarget::OutputSpendingFee,
+        AtomicU32::new(MIN_FEERATE),
+    );
+    fees
```
zoedberg: This is all duplicated code from `bitcoind.rs`. Please add a `ldk_chain_backend` directory, move `bitcoind.rs` and `indexer.rs` there (renaming them to `block_sync.rs` and `transaction_sync.rs`), and put the shared logic in the new `mod.rs` file (also move any object related to the chain sync that you added to `ldk.rs`, like `ChainBackend`). 

Generally speaking, please look for other duplicated code introduced by this PR.

src/test/mod.rs
```diff
@@ -270,6 +274,21 @@ async fn start_node(
...
+    start_node_with_sync(
+        node_test_dir,
+        node_peer_port,
+        keep_node_dir,
+        block_sync_source(),
+    )
+    .await
+}
+
+async fn start_node_with_sync(
```
zoedberg:
```suggestion
async fn start_node_with(
```
for symmetry with other new methods (e.g. `unlock_res_with`)

Cargo.toml
```diff
@@ -7,6 +7,19 @@ rust-version = "1.88.0"
...
+default = ["block-sync", "transaction-sync"]
+# Sync LDK by pulling full blocks from a trusted/local bitcoind over JSON-RPC
+# (lightning-block-sync). Stronger trust model, requires bitcoind RPC access.
+block-sync = ["dep:lightning-block-sync"]
+# Sync LDK through the configured electrum/esplora indexer
+# (lightning-transaction-sync). Lighter deployment, no bitcoind required.
```
zoedberg:
```suggestion
default = [
    "block-sync",
    "transaction-sync",
]
block-sync = [
    "dep:lightning-block-sync",
]
```

zoedberg (changes requested):

bitwalt:
Thanks again @zoedberg — went through this round and pushed an updated (still signed) commit. Per thread:

### `ldk_chain_backend` module + dedup
- Added `src/ldk_chain_backend/` with `block_sync.rs` (was `bitcoind.rs`) and `transaction_sync.rs` (was `indexer.rs`); `mod.rs` now holds the chain-backend objects moved out of `ldk.rs` (`ChainBackend`, `ChainSetup`, `DynFeeEstimator`/`DynBroadcaster`, `sync_chain_data`) plus the shared fee logic that was duplicated across the two backends — `default_fee_buckets`, the `FeeEstimator` body (`fee_from_bucket`) and the fee-estimate store loop (`store_fee_estimates`, which was triplicated).

### `routes.rs` / naming
- Dropped `rename_all` — the mode values are now `BlockSync` / `TransactionSync`, matching the existing style (`JsonRpc`, `Nia`).
- Renamed the field (and every helper/param) `chain_source` → `ldk_chain_sync`; `start_node_with_sync` → `start_node_with`.

### `indexer.rs` (now `transaction_sync.rs`)
- Dropped the `chain_hash` check and removed the unused `impl UtxoLookup for IndexerClient`; the lookup logic now lives directly in `IndexerGossipVerifier::get_utxo` (and the now-unused `network` field is gone).

### `error.rs`
- `FailedBitcoindConnection` and `NetworkMismatch` are now `#[cfg(feature = "block-sync")]` instead of `allow(dead_code)`.

### `openapi.yaml`
- `ldk_chain_sync` now uses `oneOf` + `discriminator`/`mapping` (on `mode`), like `Assignment`.

### `Cargo.toml` / CI
- Multi-line `[features]`; the single-feature build job is in place.

### `README.md`
- Moved the Sync-modes section down between `Use` and `Test`, dropped the inline JSON example (openapi covers it), switched to `cargo install ... --no-default-features --features <backend>`, and added a cross-reference from `Install`.

### Tests
- Gated with `transaction-sync` (module-level `#[cfg]`).
- Extended to a 3-node `node1 -> node2 -> node3` topology with two announced asset channels; node1 makes a multihop RGB payment to node3, which (no route hint for the public far channel) forces node1 to resolve the route from gossip and verify `node2 -> node3`'s funding output through the indexer. Runs against both electrum and esplora, then restarts all nodes (re-establishing via the indexer) and cooperatively closes a channel. Both `transaction_sync_electrum` and `transaction_sync_esplora` pass against a real regtest bitcoind + electrs + esplora.

  One note: I kept a single cooperative close in the test. With two back-to-back coop closes, the second one (peer-initiated, on the node that had just processed the first close) didn't settle its returned on-chain balance within the wait under transaction-sync — the multihop payment itself (what covers the gossip lookup) and a single close both settle fine. Happy to dig into the double-close behaviour separately if you think it's worth a closer look.

Verification: `fmt` + `clippy -D warnings` clean on default / block-sync-only / transaction-sync-only; no-feature build fails with the `compile_error!`; both integration tests green.

zoedberg (changes requested):
@bitwalt please double check your changes and my requests, there are some unaddressed things

zoedberg:
@bitwalt could you please rebase this PR on the updated master tip? Also please check my previous change requests, some were unaddressed. And finally note that we decided to drop the default indexer (see https://github.com/RGB-Tools/rgb-lightning-node/pull/147), so this PR will need to take that into account.

zoedberg:
Thanks for sticking with this.

I pushed a commit on top of yours instead of sending another round of comments, because most of it needed digging that was quicker to just do.

I found out why `ElectrumSyncClient::sync` was not confirming the funding transaction. LDK ignores the `script_pubkey` you pass to `Filter::register_tx`, and instead looks up the history of the transaction's first output. In RGB transactions that first output is the `opret` commitment, an `OP_RETURN`, and electrum servers don't index those, so the history comes back empty and the transaction never confirms. BDK hit exactly the same bug, which I fixed in bitcoindevkit/bdk#2195.

So the fix belongs in `rust-lightning`, not here. I patched our fork in https://github.com/RGB-Tools/rust-lightning/pull/37 and dropped `sync_electrum_registered_txs` completely. Your workaround was compensating for a real bug, it just wasn't at the right layer. There's a new test that reproduces it in a few seconds, so a future LDK bump can't bring it back.

Other changes:

- the gossip lookup now checks that the funding output is unspent, like the block-sync path already did. Without it a node accepts replayed announcements of channels that have already been closed
- `indexer_url` in the `TransactionSync` config is no longer optional, since master dropped the default indexer
- the single-feature builds moved from `build.yaml` into `lint.yaml`, as a matrix over all 9 valid feature combinations
- the test suite no longer needs `block-sync` to compile
- `mine()` now waits for esplora too, but only in the test that syncs through it
- a few small cleanups

None of this changes the design you landed on, it all sits on top. @bitwalt please have a look and let me know if it looks good to you.

bitwalt:
Hi Zoe, thanks for the clarification and for completing this. LGTM

Issue #61, opened 2025-06-10
add API to decode a swapstring


zoedberg:
Add a new API to decode a swapstring.

- endpoint: `/decodeswapstring`
- method: `POST` (needed to pass the parameter in a request body)
- request body: `DecodeSwapstringRequest { swapstring: String }` (the swapstring to decode)
- response body:
```rust
pub(crate) struct DecodeSwapstringResponse {
    pub(crate) qty_from: u64,
    pub(crate) qty_to: u64,
    pub(crate) from_asset: Option<String>,
    pub(crate) to_asset: Option<String>,
    pub(crate) expiry: u64,
    pub(crate) payment_hash: String,
}
```
- documentation: add to `README.md` and `openapi.yaml`
- hint: check `SwapString::from_str`

moakilodash:
i will resolve this issue.

zoedberg:
done in #128

PR #128, opened 2026-05-18
feat: add api to decode a swapstring

moakilodash:
resolving #61
let me know if i have to change anything

### DIFF
--- a/src/routes.rs
+++ b/src/routes.rs
@@ -522,6 +522,21 @@ pub(crate) struct DecodeRGBInvoiceResponse {
 }
 
 #[derive(Deserialize, Serialize)]
+pub(crate) struct DecodeSwapstringRequest {
+    pub(crate) swapstring: String,
+}
+
+#[derive(Deserialize, Serialize)]
+pub(crate) struct DecodeSwapstringResponse {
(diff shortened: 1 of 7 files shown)


src/routes.rs
```diff
@@ -1736,6 +1751,23 @@ pub(crate) async fn decode_rgb_invoice(
     }))
 }
 
+pub(crate) async fn decode_swap_string(
```
```suggestion
pub(crate) async fn decode_swapstring(
```

src/test/issue.rs
```diff
@@ -49,6 +49,48 @@ async fn issue() {
...
+        .await
+        .unwrap();
+    check_response_is_nok(
+        res,
+        reqwest::StatusCode::BAD_REQUEST,
+        "Invalid swap string 'not_a_valid_swapstring'",
+        "InvalidSwapString",
+    )
+    .await;
+
```
zoedberg: the `issue` test doesn't seem the right place for these checks, please move them to the `swap_roundtrip_assets` test

zoedberg (changes requested):
Thanks! Requested just a couple of small changes

zoedberg (approved):
LGTM, thank you!

Issue #132, opened 2026-07-15
Force close with a pending RGB HTLC permanently breaks the output sweeper (funds unclaimable)


dcorral:
If a channel is force-closed while an RGB HTLC is still pending, the commitment transaction that confirms on-chain carries the asset HTLC. When the `OutputSweeper` later tries to sweep the closer's or the counterparty's balance, `RgbOutputSpender::spend_spendable_outputs` colors the sweep transaction with the stored `transfer_info` amount, which includes the pending HTLC amount — more than the swept output actually holds. rgb-lib rejects the coloring with `InvalidColoringInfo`, the `.unwrap()` on that call panics inside the sweeper's locked section, and the poisoned lock makes **every subsequent sweep of any channel panic as well**. From that point the node can no longer claim anything — neither the RGB assets nor the plain BTC — and a restart does not help (the same tracked output is retried and panics again).

Existing force-close tests (`close_force_standard`, `close_force_other_side`, …) don't catch this because they always close channels with no HTLC in flight.

## Environment

- `rgb-lightning-node` master @ `7597bbd` ("complete listunspents and refresh APIs")
- `rust-lightning` submodule @ `5ff0ffce8` (current `rgb` branch HEAD)
- regtest via `./regtest.sh start`, Linux

## Reproduction

Test file (drop in as `src/test/force_close_pending_htlc_sweep.rs`, register `mod force_close_pending_htlc_sweep;` in `src/test/mod.rs`, run with `cargo test force_close_pending_htlc_sweep -- --test-threads=1`): see attachment or the full test at the end of this issue.

The scenario:

1. node1 issues an NIA asset and opens a colored channel to node2 (100k sat capacity, 50M msat pushed, 600 asset units).
2. A same-channel swap is set up: node1 (maker) offers 10 asset units for 25000 msat. `maker_execute` starts the circular payment maker → taker → maker; RGB commitment rounds are slow enough that the asset HTLC stays pending for a couple of seconds.
3. node2 (taker) force-closes immediately after `maker_execute`, so its commitment — carrying the pending 10-unit asset HTLC — is broadcast and confirmed.
4. The test then mines and polls: node1 must recover its ~47k sat `to_remote` and its asset balance. It never does.

## Observed behavior

Both nodes panic on their first sweep attempt and the sweeper is dead from then on:

```text
thread 'tokio-rt-worker' panicked at src/ldk.rs:1670:14:
[...]

PR #133, opened 2026-07-15
add test reproducing sweeper breakage on force close with pending RGB HTLC

dcorral:
Failing test reproducing #132

### DIFF
--- a/src/ldk.rs
+++ b/src/ldk.rs
@@ -119,6 +119,39 @@ const VANILLA_SYNC_LOOKBACK: u32 = 20;
 #[cfg(test)]
 pub(crate) static IGNORE_INBOUND_CHANNELS_ON_NODE: Mutex<Option<PublicKey>> = Mutex::new(None);
 
+/// Test-only: the node with this pubkey holds incoming payments instead of
+/// claiming them, keeping their HTLCs pending.
+#[cfg(test)]
+pub(crate) static HOLD_PAYMENT_CLAIMABLE_ON_NODE: Mutex<Option<PublicKey>> = Mutex::new(None);
+
+/// Test-only: number of payments held via [`HOLD_PAYMENT_CLAIMABLE_ON_NODE`].
+#[cfg(test)]
+pub(crate) static HELD_PAYMENT_CLAIMABLE_COUNT: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+
+/// Sets [`HOLD_PAYMENT_CLAIMABLE_ON_NODE`] and clears it on drop, so a
+/// panicking test cannot leak the hold into the next one.
+#[cfg(test)]
+pub(crate) struct HoldPaymentClaimableGuard;
+
+#[cfg(test)]
+impl HoldPaymentClaimableGuard {
+    pub(crate) fn set(node_id: PublicKey) -> Self {
+        HELD_PAYMENT_CLAIMABLE_COUNT.store(0, std::sync::atomic::Ordering::SeqCst);
+        *HOLD_PAYMENT_CLAIMABLE_ON_NODE.lock().unwrap() = Some(node_id);
+        Self
+    }
+}
+
+#[cfg(test)]
+impl Drop for HoldPaymentClaimableGuard {
+    fn drop(&mut self) {
+        if let Ok(mut hold) = HOLD_PAYMENT_CLAIMABLE_ON_NODE.lock() {
+            *hold = None;
+        }
+    }
+}
+
 pub(crate) struct LdkBackgroundServices {
     stop_processing: Arc<AtomicBool>,
     peer_manager: Arc<PeerManager>,
@@ -978,6 +1011,17 @@ async fn handle_ldk_events(
                 payment_hash,
                 amount_msat,
             );
+            #[cfg(test)]
+            if HOLD_PAYMENT_CLAIMABLE_ON_NODE
+                .lock()
+                .unwrap()
+                .as_ref()
+                .is_some_and(|id| *id == unlocked_state.channel_manager.get_our_node_id())
+            {
+                tracing::info!("TEST: holding PaymentClaimable for {}", payment_hash);
+                HELD_PAYMENT_CLAIMABLE_COUNT.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+                return Ok(());
+            }
             let payment_preimage = match purpose {
                 PaymentPurpose::Bolt11InvoicePayment {
                     payment_preimage, ..
--- a/src/ldk.rs
+++ b/src/ldk.rs
@@ -84,6 +84,8 @@ use std::net::ToSocketAddrs;
 use std::net::{SocketAddr, TcpListener};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+#[cfg(test)]
+use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 use std::time::{Duration, SystemTime};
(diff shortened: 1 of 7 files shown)


zoedberg:
Thanks for the reproduction, it pins down a real "funds unclaimable" bug and closes a genuine coverage gap (none of our existing force-close tests exercise a sweep with a pending HTLC).

The main thing that needs addressing before we can start discussing the fix is determinism. Right now the test relies on force-closing quickly enough after `maker_execute` to catch the asset HTLC still pending, but nothing actually guarantees that's the case at close time. If the commitment rounds happen to settle first, the HTLC is gone, the bug doesn't reproduce, and the test passes for the wrong reason, a silent false green.

Could you make the reproduction deterministic by asserting the pending HTLC was genuinely present on the confirmed commitment before checking recovery? That way a green result actually means "swept successfully despite a pending HTLC" rather than "the HTLC happened to clear in time."

A smaller note: the comment says "both nodes must recover" but the loop only polls node1. Please assert also node2's recovery.

dcorral:
The test now verifies on-chain that the confirmed commitment actually contains the pending HTLC output before checking recovery, if the HTLC settled before the close, that attempt is discarded and retried on a fresh channel (up to 5 times), and if none is caught the test fails with a distinct error instead of passing. So a green run can only mean "swept successfully despite a pending HTLC", never a lucky race. 

Also added the node2 recovery assertions. 

We went with on-chain verification rather than a hold-HTLC test hook to keep the change self-contained (no rust-lightning changes)

zoedberg:
I don't think the retry loop makes the test deterministic; it just raises the probability of catching the race.

The core issue: with a "try up to 5 times, fail if none caught the HTLC" design, you've correctly eliminated the false green (a pass can now only mean the bug path was exercised). But you've converted it into a false red / flake: if all 5 attempts happen to settle the HTLC before the close the test fails even though the production code is perfectly fine. That's a flaky test, and an intermittently-red regression guard tends to get muted or ignored over time, which defeats the purpose. Bumping 5 to 20 just moves the probability around; it never reaches 0 or 1. It's still timing.

Deterministic means every run confirms a commitment carrying the pending HTLC, by construction. That requires a test hook that holds the HTLC unsettled (or forces the commitment to be signed while it's still pending), so the on-chain state is guaranteed rather than sampled.

Could you switch to a mock/hook-based approach that deterministically holds the HTLC pending across the force-close?

dcorral:
Added a test-only HOLD_PAYMENT_CLAIMABLE_ON_NODE static, mirroring IGNORE_INBOUND_CHANNELS_ON_NODE. The payee's event handler holds the incoming payment instead of claiming, so the HTLC is pending by construction and the on-chain "commitment carries the HTLC" check is now a hard assertion that holds on every run. No timing anywhere.

zoedberg:
@gofman8 now that the test is deterministic I can start working on a solution, since in another communication channel you mentioned a possible solution please share it here

dcorral:
Proposed fix, in three parts:

**1. Never panic in the sweep path.** Today a failed sweep panics while the sweeper's lock is held, poisoning it, after that the node can't claim anything (assets or BTC, any channel) and a restart doesn't help. Replace the unwraps in `RgbOutputSpender::spend_spendable_outputs` with a logged error so the sweeper just retries on its next tick.

**2. Color the sweep with the amount actually on the swept output.** The rejection happens because the sweep is colored with the stored `transfer_info.rgb_amount`, which is the total claimable from the close *including* the pending HTLC, but the HTLC amount sits on its own output. Instead, query the exact amount assigned to the output being spent and color with that. Needs a small rgb-lib addition, e.g. `get_outpoint_fungible_assignments(contract_id, outpoint) -> u64` (a thin wrapper over the internal `contract_assignments_for`).

With 1 + 2 both nodes recover their BTC and the channel-balance assets even with a payment in flight, the test in this PR goes green.

**3. Follow-up: the in-flight amount itself.** The HTLC's assets are claimed on-chain later (preimage or timeout), but the claim transaction gets no `transfer_info`, so those assets are currently lost even after the claim confirms. Fixing that means coloring/recording the HTLC claim tx in rust-lightning, independent of 1 + 2, can land separately.

Thoughts?

src/test/force_close_pending_htlc_sweep.rs
```diff
@@ -0,0 +1,176 @@
+use std::sync::atomic::Ordering;
+
+use crate::ldk::{HELD_PAYMENT_CLAIMABLE_COUNT, HOLD_PAYMENT_CLAIMABLE_ON_NODE};
+
+use super::*;
+
+const TEST_DIR_BASE: &str = "tmp/force_close_pending_htlc_sweep/";
+
+async fn spendable_sats(node_address: SocketAddr) -> u64 {
```
zoedberg: please move the utility methods here to `test/mod.rs`, deduplicating where possible (e.g. refresh call)

src/test/force_close_pending_htlc_sweep.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// Force close with a pending RGB HTLC: both nodes must still recover their
+/// BTC and node1 its assets. The HTLC is held pending by construction: node2
+/// is set to hold incoming payments (HOLD_PAYMENT_CLAIMABLE_ON_NODE), so the
+/// commitment provably carries the asset HTLC when node2 force-closes.
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[traced_test]
+async fn force_close_pending_htlc_sweep() {
```
zoedberg: please rename it to `close_force_pending_htlc` (and its directory accordingly)

src/test/force_close_pending_htlc_sweep.rs
```diff
@@ -0,0 +1,176 @@
...
+            node1_assets_ok = true;
+        }
+        println!(
+            "recovery: node1_btc_ok={node1_btc_ok} node2_btc_ok={node2_btc_ok} node1_assets={node1_assets}"
+        );
+        if node1_btc_ok && node2_btc_ok && node1_assets_ok {
+            break;
+        }
+    }
+    *HOLD_PAYMENT_CLAIMABLE_ON_NODE.lock().unwrap() = None;
```
zoedberg: `HOLD_PAYMENT_CLAIMABLE_ON_NODE` is only reset after the recovery loop, so the earlier panics leave it set. The next test then silently gets a node that never claims payments, please make it a `Drop` guard

src/test/force_close_pending_htlc_sweep.rs
```diff
@@ -0,0 +1,176 @@
...
+        refresh_transfers_tolerant(node1_addr).await;
+        refresh_transfers_tolerant(node2_addr).await;
+        if !node1_btc_ok && spendable_sats(node1_addr).await > node1_spendable_before + 30_000 {
+            node1_btc_ok = true;
+        }
+        if !node2_btc_ok && spendable_sats(node2_addr).await > node2_spendable_before + 30_000 {
+            node2_btc_ok = true;
+        }
+        let node1_assets = asset_balance_spendable(node1_addr, &asset_id).await;
+        if node1_assets >= 990 {
```
zoedberg: let's test it's exactly 990, we'll eventually update the test after addressing the other bug mentioned in part 3

src/test/force_close_pending_htlc_sweep.rs
```diff
@@ -0,0 +1,176 @@
...
+fn node_ldk_logs(node_test_dir: &str) -> String {
+    let log_path = Path::new(node_test_dir)
+        .join(LDK_DIR)
+        .join(crate::utils::LOGS_DIR)
+        .join(LDK_LOGS_FILE);
+    std::fs::read_to_string(log_path).unwrap_or_default()
+}
+
+/// Returns the txid of the confirmed commitment for the given channel.
+async fn confirmed_commitment_txid(node_test_dir: &str, channel_id: &str) -> String {
```
zoedberg: this does `logs[pos + needle.len() .. pos + needle.len() + 64]` on a `read_to_string`, which panics with a byte-index out of range if the log is caught mid-write. Use `BufReader::new(file).lines()` as the other tests reading logs are doing. Actually if you don't mind I think this is a good time to add a `ldk_log_lines(node_test_dir)` method to dedup the "find a log" logic scattered around the tests

zoedberg (changes requested):
1. **Never panic in the sweep path.** You're right that the poisoning is what turns this into "can't claim anything, restart doesn't help". But with no bugs your argument only applies to failures that legitimately can happen (e.g. `get_tx_height`). I fixed this by handling transient errors separately, making later retries possible. Everything else shouldn't happen at all, so I kept it fatal, which makes easier to notice bugs. At this stage I think it's more important to blow up immediately rather than get logged once a block and retried forever (a bug that slides silently is worse than one that stops the node)

2. **Color the sweep with the amount actually on the swept output.** This is the actual bug. But the solution is much more straightforward than the one you proposed: `color_commitment` already computes the correct per-vout map, we just need to save that instead of the summed amount. Deriving the input amount from the stock rather than from what we actually colored by using a `get_outpoint_fungible_assignments`  method would be wrong. Those are two sources that have to agree and if they ever drift we're back to `InvalidColoringInfo` (or some other error) for a subtler reason that would be much harder to debug than this one. Coloring the spend with the spent transaction's own map makes the mismatch impossible by construction rather than merely unlikely and it needs no rgb-lib changes

3. **Follow-up: the in-flight amount itself.** `color_htlc` does write a `TransferInfo`, but only when we force-close ourselves and claim through a second-stage HTLC transaction. Here node2 closed, so node1 claims the HTLC directly from node2's commitment and that path never colors anything. That's the gap. I think this requires a separate PR where we extend this test, could you please do that? Then I can work on that fix too

Please revert your changes to the CI, address the requested changes and rebase this commit on top of the updated `master` tip, I'll then push my fix on top.

dcorral:
All addressed: 
- CI reverted 
- rebased on 017cafb
- test renamed to close_force_pending_htlc
- helpers moved into test/mod.rs (with ldk_log_lines also deduplicating the log reads in refuse_high_fees, openchannel_no_indexer and swap_roundtrip_fail_whitelist)
- the hold hook is now set through a Drop guard so a panicking test can't leak it
- the asset check is exactly 990. 

Ready for your fix on top

Issue #32, opened 2024-07-09
Broking RLN API on /sendpayment though routing node with no in/out liquidity 


gofman8:
During our work on RGB OS/Thunderstack Cloud we faced an issue with RLN API critical error. Below we're attaching exact steps to reproduce the issue.  Next steps expects to have up and running and initialized a 3 nodes: A,B,C. 

### Send BTC to Nodes
Run the following command on your server:
```bash
./regtest.sh sendtoaddress <address> <amount>
```
Example:
```bash
./regtest.sh sendtoaddress bcrt1qnc5y6j6dmejrkwy93farhvpezk0lf46gk7aecs 10
```
then

```bash
./regtest.sh mine 10
```
do it for A, B, C

# Node B:
### 1.1 Create Asset

Call the `/createutxos` endpoint:
```json
{
  "up_to": false,
  "num": 4
}
```

Call the `/issueassetnia` endpoint with the following example body:
```json
{
  "amounts": [200],
  "ticker": "USDT",
  "name": "Tether",
  "precision": 0
}
```

rgb:2RCNGGK-SPo4jJYWx-uBR5XwQd4-mUAfzyten-gzxMLo1vG-7Ktxsxp

/assetbalance node B
 ```json 
 {
  "settled": 200,
  "future": 200,
  "spendable": 200,
  "offchain_outbound": 0,
  "offchain_inbound": 0
}
 ```
### 2. Open channel with Node A with 50 tokens (Channels API)
/assetbalance balance Node B
 ```json 
 {
  "settled": 200,
  "future": 150,
  "spendable": 0,
  "offchain_outbound": 50,
  "offchain_inbound": 0
}
 ```
after mine 
 ```json 
    {
      "settled": 150,
      "future": 150,
      "spendable": 150,
      "offchain_outbound": 50,
      "offchain_inbound": 0
    }
```
/assetbalance balance Node A
 ```json 
    {
      "settled": 0,
      "future": 0,
      "spendable": 0,
      "offchain_outbound": 0,
      "offchain_inbound": 50
    }
 ```
### 3. Open channel with Node C with 50 tokens (Channels API)
/assetbalance balance Node B
 ```json 
{
  "settled": 150,
  "future": 150,
  "spendable": 150,
  "offchain_outbound": 100,
  "offchain_inbound": 0
}
 ```
after mine 
 ```json 
{
  "settled": 100,
  "future": 100,
  "spendable": 100,
  "offchain_outbound": 100,
[...]

PR #136, opened 2026-07-24
Restart the payer with a pending RGB HTLC in close_force_pending_htlc

dcorral:
Restarts the payer in `close_force_pending_htlc` while the colored HTLC is still pending, so the channel state must round-trip through persistence with an HTLC in flight. Without RGB-Tools/rust-lightning#32 the node fails to come back up at this point, with it the test passes. The rust-lightning submodule is updated to include that fix.

### DIFF
--- a/src/test/close_force_pending_htlc.rs
+++ b/src/test/close_force_pending_htlc.rs
@@ -62,6 +62,44 @@ async fn close_force_pending_htlc() {
         tokio::time::sleep(std::time::Duration::from_secs(1)).await;
     }
 
+    // Restart the payer with the colored HTLC still pending: its channel must
+    // round-trip through persistence with the HTLC in flight.
+    shutdown(&[node1_addr]).await;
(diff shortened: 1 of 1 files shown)


zoedberg (changes requested):
Please first push only the test and let the CI run so that we see that the test reproduces the bug. Then add the commit that updates the submodule containing the fix

zoedberg:
@dcorral I noticed the fix is for both inbound and outbound HTLCs, but the test is restarting only one node. Could you also add a node2 restart (before the force-close, mirroring the node1 one) so the test exercises the inbound deserialization path too, not just outbound? That way the regression covers both write sites the fix touches.

dcorral:
Done, the payee is now restarted as well before the force close, so both the outbound and inbound pending-HTLC paths go through deserialization. Kept the PR test-only for now so CI shows the failure; the submodule update with the fix follows once the red run is on record.

zoedberg (approved):
LGTM, thank you!

PR #137, opened 2026-07-28
Network serialization round-trip fails for mainnet and signetcustom

dcorral:
`WalletSource::list_confirmed_utxos` converts the wallet network by serializing it to a string and parsing it back: `Network::from_str(&self.bitcoin_network().to_string().to_lowercase()).unwrap()`. `rgb_lib::BitcoinNetwork`'s `Display` delegates to `Debug`, so `Mainnet` serializes as `"mainnet"` and `SignetCustom` as `"signetcustom"`, neither of which `bitcoin::Network::from_str` accepts, and the `.unwrap()` panics deterministically the first time `Event::BumpTransaction` fires on those networks (anchor CPFP: force closes, HTLC resolution). Since the panic happens inside the event-handler future, it also kills the spawned `process_events_async` task, leaving the node running without event processing.

The added unit test runs the same round-trip for every `BitcoinNetwork` variant and fails on `Mainnet` (`ParseNetworkError("mainnet")`); with that variant excluded it fails on `SignetCustom` too. The existing suite never catches this because it runs on regtest, whose network name parses back fine.

### DIFF
--- a/src/rgb.rs
+++ b/src/rgb.rs
@@ -885,3 +885,25 @@ pub(crate) fn get_rgb_channel_info_optional(
     let rgb_info = parse_rgb_channel_info(&info_file_path);
     Some((rgb_info, info_file_path))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bitcoin_network_serialization_round_trips() {
+        for network in [
+            BitcoinNetwork::Mainnet,
+            BitcoinNetwork::Testnet,
+            BitcoinNetwork::Testnet4,
+            BitcoinNetwork::Signet,
+            BitcoinNetwork::Regtest,
+            BitcoinNetwork::SignetCustom,
+        ] {
+            assert!(
+                Network::from_str(&network.to_string().to_lowercase()).is_ok(),
+                "{network:?} does not parse back"
+            );
+        }
+    }
+}


dcorral:
Proposed fix is dropping the string round-trip in favor of the `From<BitcoinNetwork>` conversion rgb-lib already provides, which covers every variant and cannot fail.

This was already implemented in the UTEXO fork (https://github.com/UTEXO-Protocol/rgb-lightning-node/pull/113)

dcorral:
One thing worth flagging beyond the serialization fix itself. The panic happens inside the event handler future, and process_events_async runs in a tokio::spawned task whose JoinHandle is only awaited at shutdown. So when the handler panics, only the background processor dies: the HTTP API, peer connections and chain monitoring all keep running. The node looks healthy from the outside while event processing and channel manager persistence are dead, and its state silently diverges for as long as it stays up. We saw a node run for two days in this condition.

In our fork we solved it by supervising the background processor task: if it ever terminates unexpectedly (panic or early return) the process logs the cause and exits with a distinct code, so an orchestrator restarts it and the failure is loud instead of silent (https://github.com/UTEXO-Protocol/rgb-lightning-node/pull/113).

zoedberg:
I assume the PR is just to show that coverting `BitcoinNetwork` to `Network` using string conversion is incorrect (not an actual roundtrip test), so closing the PR. But thanks for finding the bug in `WalletSource::list_confirmed_utxos`. I agree the fix is using the already existing conversion (`let network: Network = self.bitcoin_network().into();` instead of `Network::from_str(&self.bitcoin_network().to_string().to_lowercase()).unwrap();`).

About the event processing panic, terminating the node when the background processor dies is the right thing to do. The node surfing a crash with no apparent changes from the user PoV is dangerous.
Automatically restarting the node is not appropriate though: a panic means a bug and at RLN's current stage a bug should be an incident someone reports, not something infrastructure absorbs.

Issue #33, opened 2024-07-12
Channels disappearing after paying invoice (no errors in logs)


gofman8:
### Run nodes A,B,C 
Run nodes A,B,C call `/init` & `/unlock` for each node.

### Send BTC to Nodes
Run the following command on your server:
```bash
./regtest.sh sendtoaddress <address> <amount>
```
Example:
```bash
./regtest.sh sendtoaddress bcrt1qnc5y6j6dmejrkwy93farhvpezk0lf46gk7aecs 10
```
then

```bash
./regtest.sh mine 10
```
do it for A, B, C

# Node B:
### 1.1 Create Asset

Call the `/createutxos` endpoint:
```json
{
  "up_to": false,
  "num": 4
}
```

Call the `/issueassetnia` endpoint with the following example body:
```json
{
  "amounts": [200],
  "ticker": "USDT",
  "name": "Tether",
  "precision": 0
}
```
- On Node B and open channel to Node A and C (50 assets) ->  channels successfully created and `/listchannels` returns 2 channel items

- /keysend 25 assets from Node B to Node A -> payment success

 | Action                                          |   A set |   A fut |   A spen |   A out |   A in |   B set |   B fut |   B spen |   B out |   B in |   C set |   C fut |   C spen |   C out |   C in |
|:------------------------------------------------|--------:|--------:|---------:|--------:|-------:|--------:|--------:|---------:|--------:|-------:|--------:|--------:|---------:|--------:|-------:|
| 1. Issue 200 assets on Node B and open channel to Node A and C |       0 |       0 |        0 |       0 |     50 |     100 |     100 |      100 |     100 |      0 |       0 |       0 |        0 |       0 |     50 |
| 2. /keysend 25 assets to Node A                 |       0 |       0 |        0 |      25 |     25 |     100 |     100 |      100 |      75 |     25 |       0 |       0 |        0 |       0 |     50 |

Node C calls `/lninvoice` (25 assets)
**Request:**
```json
{
  "amt_msat": 3000000,
  "expiry_sec": 620,
  "asset_id": "rgb:2Fy9jVd-4BCGgtPXi-wFcrAHqrD-ADBQQBKbt-cDKA7LA5y-wiuHGJT",
  "asset_amount": 25
}
```
**Response:**
```json
{
[...]

PR #138, opened 2026-07-28
Add test for push_asset_amount underflow DoS on funding acceptor

0xaudron:
The acceptor computed remote_rgb_amount = channel_rgb_amount - push_amount without checking push_amount <= channel_rgb_amount, so a counterparty sending a larger value underflowed it (release builds have overflow checks off), aborting RGB coloring and panicking the acceptor's event handler.

The REST layer clamps push_asset_amount, so reaching this needs a modified wire client; FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE models that by making the initiator emit an out-of-range value on the wire. The test asserts the acceptor rejects the funding and stays responsive.

Bumps rust-lightning to the commit rejecting out-of-range push_asset_amount in handle_funding.

### DIFF
--- a/src/ldk.rs
+++ b/src/ldk.rs
@@ -130,6 +130,12 @@ pub(crate) static HOLD_PAYMENT_CLAIMABLE_ON_NODE: Mutex<Option<PublicKey>> = Mut
 #[cfg(test)]
 pub(crate) static HELD_PAYMENT_CLAIMABLE_COUNT: AtomicUsize = AtomicUsize::new(0);
 
+// Test-only: the node with this pubkey emits a `push_asset_amount` greater than the channel asset
+// amount on the wire in `open_channel`, regardless of the value validated by its REST layer. Used
+// to model a channel counterparty whose wire client is not bound by the sender-side clamp.
+#[cfg(test)]
+pub(crate) static FORCE_PUSH_ASSET_AMOUNT_ON_NODE: Mutex<Option<PublicKey>> = Mutex::new(None);
+
 pub(crate) struct LdkBackgroundServices {
     stop_processing: Arc<AtomicBool>,
     peer_manager: Arc<PeerManager>,
--- a/src/routes.rs
+++ b/src/routes.rs
@@ -79,6 +79,8 @@ use tokio::{
     sync::MutexGuard as TokioMutexGuard,
 };
 
+#[cfg(test)]
+use crate::ldk::FORCE_PUSH_ASSET_AMOUNT_ON_NODE;
 use crate::swap::{SwapData, SwapInfo, SwapString};
 use crate::utils::{
     check_already_initialized, check_channel_id, check_password_strength, check_password_validity,
@@ -3495,7 +3497,16 @@ pub(crate) async fn open_channel(
             let schema = unlocked_state
                 .rgb_get_asset_metadata(*contract_id)?
                 .asset_schema;
-            (Some((*contract_id, payload.push_asset_amount)), Some(schema))
+            #[cfg(not(test))]
+            let wire_push_asset_amount = payload.push_asset_amount;
+            #[cfg(test)]
+            let wire_push_asset_amount = match *FORCE_PUSH_ASSET_AMOUNT_ON_NODE.lock().unwrap() {
+                Some(node) if node == unlocked_state.channel_manager.get_our_node_id() => {
+                    Some(*asset_amount + 1)
+                }
+                _ => payload.push_asset_amount,
+            };
+            (Some((*contract_id, wire_push_asset_amount)), Some(schema))
         } else {
             let balance = unlocked_state.rgb_get_btc_balance(true)?;
             if payload.capacity_sat > balance.vanilla.spendable {
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -29,8 +29,8 @@ use tracing_test::traced_test;
 use crate::disk::LDK_LOGS_FILE;
 use crate::error::APIErrorResponse;
 use crate::ldk::{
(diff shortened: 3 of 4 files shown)


src/ldk.rs
```diff
@@ -130,6 +130,13 @@ pub(crate) static HOLD_PAYMENT_CLAIMABLE_ON_NODE: Mutex<Option<PublicKey>> = Mut
 #[cfg(test)]
 pub(crate) static HELD_PAYMENT_CLAIMABLE_COUNT: AtomicUsize = AtomicUsize::new(0);
 
+// Test-only: the node with this pubkey emits the given `push_asset_amount` on the wire in
+// `open_channel`, regardless of the value validated by its REST layer. Used to model a channel
+// counterparty whose wire client is not bound by the sender-side clamp in `open_channel`.
+#[cfg(test)]
+pub(crate) static FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE: Mutex<Option<(PublicKey, u64)>> =
```
zoedberg: I don't see the use case of specifying a specific push asset amount, I think this variable should be similar to `IGNORE_INBOUND_CHANNELS_ON_NODE`, where you just turn on/off a behavior. The push amount could be hardcoded to a very high value when `FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE` is set. This way you could use the `NodeOverrideGuard` and remove the `Drop` logic you added in `src/test/push_asset_amount_underflow.rs` 

I also propose to rename the variable to `FORCE_PUSH_ASSET_AMOUNT_ON_NODE`

src/routes.rs
```diff
@@ -3495,7 +3495,25 @@ pub(crate) async fn open_channel(
...
                 .asset_schema;
-            (Some((*contract_id, payload.push_asset_amount)), Some(schema))
+            // The value placed on the wire in `open_channel`. Production always uses the
+            // REST-validated `push_asset_amount`. In tests a node may be told to emit a different
+            // value (see `FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE`), modelling a counterparty whose
+            // wire client is not bound by the sender-side clamp above.
+            #[cfg(not(test))]
+            let wire_push_asset_amount = payload.push_asset_amount;
+            #[cfg(test)]
+            let wire_push_asset_amount = match *crate::ldk::FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE
```
```suggestion
            let wire_push_asset_amount = match *FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE
```

src/test/push_asset_amount_underflow.rs
```diff
@@ -0,0 +1,83 @@
+use super::*;
+
+const TEST_DIR_BASE: &str = "tmp/push_asset_amount_underflow/";
```
zoedberg:
```suggestion
const TEST_DIR_BASE: &str = "tmp/push_asset_amount_above_chan_amt/";
```
please rename the file and test name too

src/test/push_asset_amount_underflow.rs
```diff
@@ -0,0 +1,83 @@
...
+    let node2_pubkey = node_info(node2_addr).await.pubkey;
+
+    let asset_id = issue_asset_nia(node1_addr).await.asset_id;
+
+    connect_peer(
+        node1_addr,
+        &node2_pubkey,
+        &format!("127.0.0.1:{NODE2_PEER_PORT}"),
+    )
+    .await;
```
zoedberg: this should not be necessary, `/openchannel` already connects the peers

src/routes.rs
```diff
@@ -3495,7 +3495,25 @@ pub(crate) async fn open_channel(
             let schema = unlocked_state
                 .rgb_get_asset_metadata(*contract_id)?
                 .asset_schema;
-            (Some((*contract_id, payload.push_asset_amount)), Some(schema))
+            // The value placed on the wire in `open_channel`. Production always uses the
+            // REST-validated `push_asset_amount`. In tests a node may be told to emit a different
+            // value (see `FORCE_WIRE_PUSH_ASSET_AMOUNT_ON_NODE`), modelling a counterparty whose
+            // wire client is not bound by the sender-side clamp above.
```
```suggestion
```
explanation on top of the constant should be enough

zoedberg (changes requested):
Thank you! I requested some changes, also this branch has conflicts that must be resolved. Finally please do not commit the updated submodule pointer, I first would like to see the CI fail, then I'll push your fix on the upstream remote so you will be able to change the pointer without any change to the `.gitmodules` file

0xaudron:
Addressed all review points: `FORCE_PUSH_ASSET_AMOUNT_ON_NODE` is now a simple on/off flag like `IGNORE_INBOUND_CHANNELS_ON_NODE` and uses `NodeOverrideGuard` (custom `Drop` removed), test and file renamed to `push_asset_amount_above_chan_amt`, the redundant `connect_peer` is gone, and the explanation lives only on the constant. Rebased on master and dropped the submodule pointer. 

CI should now fail on the new test, ready for you to push the LDK fix upstream.

zoedberg:
@0xaudron the CI cannot run because the branch has still conflicts that must be resolved. You should rebase this on top of the latest master tip. Please before the rebase also squash the commits into a single one

0xaudron:
@zoedberg can you check it now ?

zoedberg:
LGTM, I'm waiting for your rebase in https://github.com/RGB-Tools/rust-lightning/pull/33 , after that I'll push the branch on the public repo so you can add a commit here that updates the submodule pointer

0xaudron:
done @zoedberg

zoedberg:
@0xaudron I pushed the branch, you can now add a commit here that updates the submodule pointer

zoedberg:
@0xaudron I will add the commit myself, so we can then proceed to merging this

PR #141, opened 2026-07-30
Reject a colored channel open with no allocation slot available

### LINKED ISSUE #140: /openchannel accepts a colored channel it cannot fund, then force-closes it
### What happens

Opening a colored channel when no colorable UTXO has a free allocation slot returns `200 OK`. The channel is negotiated with the peer, and the funding then fails inside the `FundingGenerationReady` handler, where `handle_funding_prepare_err` → `abort_funding` calls `force_close_broadcasting_latest_txn`.

The caller has already been told the open succeeded, so anything it does on the strength of that response — recording the channel, releasing a payment, reporting success upstream — is based on a channel that no longer exists. From the counterparty's side it is a force close carrying the reason string `Insufficient allocations`. Nothing reaches the chain (`broadcasted_latest_txn: Some(false)`).

### Reproducing

With one colorable UTXO taken by an issuance, so no slot is free:

```
INFO EVENT: initiated channel with peer 025795d6053ff8fc...
INFO uri=/openchannel status_code=200 OK ENDED in 19.92425ms
...
ERROR Cannot open channel: Insufficient allocations
INFO EVENT: Channel <id> closed due to: HolderForceClosed {
         broadcasted_latest_txn: Some(false), message: "Insufficient allocations" }
```

A test case is included in the PR, as an addition to `openchannel_fail`.

### Why this looks fixable up front

`open_channel` already rejects, synchronously, what it can determine from local wallet state: capacity above `balance.colored.spendable`, asset amount above `balance.spendable`. Whether a colorable UTXO has a free allocation slot is the same kind of check on the same state — `rgb_list_unspents` is already available on `UnlockedAppState` — and with `max_allocations_per_utxo: 1` it is deterministic before the peer is contacted.

rgb-lib raises `InsufficientAllocationSlots` from two places, and both need at least one free colorable UTXO: `wallet/offline.rs` when picking a UTXO for the change allocation, and `wallet/online.rs` when it needs a free UTXO as an additional BTC input after `InsufficientBitcoins`. The existing `InsufficientAllocationSlots` → `NoAvailableUtxos` mapping already produces a suitable message (`hint: call createutxos`), so no new error variant is needed.

This does not remove the need to handle the error in `FundingGenerationReady`: concurrent opens can each pass the check and then contend for the same slot. It removes the deterministic case, where the wallet has no free slot at all and every colored open is doomed before it starts.

bitwalt:
Fixes #140.

`open_channel` already rejects up front what it can determine from local wallet state — capacity above `balance.colored.spendable`, asset amount above `balance.spendable`. This adds the remaining check of the same kind: whether any colorable UTXO has a free allocation slot for the funding's change allocation.

Without it the request returns `200 OK`, the channel is negotiated with the peer, and the funding then fails in `FundingGenerationReady`, where the only way out is force-closing the channel that has just been negotiated — with the caller already told the open succeeded.

The first commit adds the case to `openchannel_fail`, at the point where the issuance has taken the node's only colorable UTXO, and fails on its own. The second makes it pass. `InsufficientAllocationSlots` already maps to `NoAvailableUtxos`, whose message (`hint: call createutxos`) is what a caller needs, so no new error variant.

This does not make the abort impossible — concurrent opens can each pass the check and then contend for the same slot, and rgb-lib can also consume a free UTXO as an extra BTC input — so `FundingGenerationReady` still has to handle the error. It removes the deterministic case, where the wallet has no free slot at all.

### DIFF
--- a/src/routes.rs
+++ b/src/routes.rs
@@ -3492,6 +3492,19 @@ pub(crate) async fn open_channel(
             if *asset_amount > balance.spendable {
                 return Err(APIError::InsufficientAssets);
             }
+            // funding the channel will need a colorable UTXO for the change allocation
+            if !unlocked_state
+                .rgb_list_unspents(false, true)?
+                .iter()
+                .any(|u| {
+                    u.utxo.colorable
+                        && u.utxo.exists
+                        && u.pending_blinded == 0
+                        && u.rgb_allocations.is_empty()
+                })
+            {
+                return Err(APIError::NoAvailableUtxos);
+            }
             let schema = unlocked_state
                 .rgb_get_asset_metadata(*contract_id)?
                 .asset_schema;
--- a/src/test/openchannel_fail.rs
+++ b/src/test/openchannel_fail.rs
@@ -21,6 +21,36 @@ async fn openchannel_fail() {
     let node2_info = node_info(node2_addr).await;
     let node2_pubkey = node2_info.pubkey;
 
+    // no colorable UTXO left for the change allocation
+    let res = open_channel_raw(
+        node1_addr,
+        &node2_pubkey,
+        Some(NODE2_PEER_PORT),
+        Some(31_000),
+        None,
+        Some(500),
+        Some(&asset_id),
+        None,
+        None,
+        None,
+        None,
+        true,
+        true,
+    )
+    .await;
+    check_response_is_nok(
+        res.unwrap_err(),
+        reqwest::StatusCode::FORBIDDEN,
+        "No uncolored UTXOs are available",
+        "NoAvailableUtxos",
+    )
+    .await;
+
+    let channels_1 = list_channels(node1_addr).await;
+    let channels_2 = list_channels(node2_addr).await;
+    assert_eq!(channels_1.len(), 0);
+    assert_eq!(channels_2.len(), 0);
+
     // insufficient BTC funds
     let res = open_channel_raw(
         node1_addr,


zoedberg:
The RGB change doesn't need a pre-existing UTXO, in rgb-lib's `get_change_seal`, when there's a BTC change output the seal goes on a vout of the funding TX itself. It only grabs an existing UTXO when the funding TX has no change output at all.

Your test hits `InsufficientAllocationSlots` for a different reason: node1's single colorable UTXO is 32k (`UTXO_SIZE_SAT`) vs a 31k channel + fee, so rgb-lib goes looking for a free UTXO to add as an extra BTC input. The real gap is that `capacity_sat > balance.colored.spendable` ignores the fee, not that no slot is free, which is the known limitation the comment just above already calls out: `checks on balances here are not precise since they do not take fees into account`.

The check you added would reject also valid channel openings: a wallet with one large colorable UTXO holding the asset and nothing free can fund the channel fine today, but would get a 403 error with your change.

When the API is called we can't know the TX size to calculate the fees in advance, so I think leaving the `FundingGenerationReady` event to close the channel is correct. In LN nodes it's normal for a successful openchannel response not to guarantee the channel opens, there are plenty of other post-response failure modes, and the UI has to handle that anyway. So I'd rather not make the checks here more precise. Closing.

bitwalt:
Thanks for the detailed correction, sorry I got the mechanism wrong. The bug was on our side, we weren't maintaining the allocation pool, so funding kept failing and every attempt ended in a force close. Agreed that's exactly the kind of post-response failure a client has to handle, not something openchannel should guarantee against.

PR #146, opened 2026-08-05
Shut down on panic or unexpected background processor exit

dcorral:
Follow-up to the discussion in #137. A panic inside the event handler (or any spawned task) kills only that task: `process_events_async` runs in a tokio spawned task whose `JoinHandle` is only awaited at shutdown, so when it dies the HTTP API, peer connections and chain monitoring keep running. The node looks healthy from the outside while event processing and channel manager persistence are dead, and it later exits with code 0.

This PR makes that failure loud, reusing the existing shutdown path:

- a `FATAL_ERROR: OnceLock<String>` in `utils.rs` records the first fatal cause
- the BP spawn is wrapped: if `process_events_async` returns while `stop_processing` was never set, the cause is logged, `FATAL_ERROR` is set and `cancel_token.cancel()` triggers the same graceful teardown as the `/shutdown` API
- a global panic hook installed in `main()` (after `app()`, capturing the cancel token) does the same for panics, which unwind through the await and would skip the wrapper; it chains the default hook to keep backtraces. Since `main()` doesn't run in the test binary, tests are unaffected
- after `axum::serve` returns, `main` exits with code 70 (sysexits `EX_SOFTWARE`) if `FATAL_ERROR` is set, so an orchestrator can tell a crash from an operator stop; this is the only `process::exit`

Note the panic hook intentionally makes a panic in *any* task fatal, not just the BP: the node surviving a crash with no visible change is the dangerous case, and a panic should surface as an incident rather than be absorbed. 

Graceful shutdowns are unchanged: `stop_processing` is set before `bp_exit` is signalled, so the wrapper stays silent and the process still exits 0.

### DIFF
--- a/src/ldk.rs
+++ b/src/ldk.rs
@@ -111,7 +111,7 @@ use crate::swap::SwapData;
 use crate::utils::{
     check_port_is_available, connect_peer_if_necessary, do_connect_peer, get_current_timestamp,
     hex_str, AppState, StaticState, UnlockedAppState, ELECTRUM_URL_MAINNET, ELECTRUM_URL_REGTEST,
-    ELECTRUM_URL_SIGNET, ELECTRUM_URL_TESTNET, ELECTRUM_URL_TESTNET4,
+    ELECTRUM_URL_SIGNET, ELECTRUM_URL_TESTNET, ELECTRUM_URL_TESTNET4, FATAL_ERROR,
 };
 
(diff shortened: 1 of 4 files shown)


zoedberg:
@dcorral Thank you! Please check the commit I added, it makes some small improvements. Let me know if they look good to you

dcorral:
All good, thanks!

PR #151, opened 2026-08-10
Make listtransfers asset_id optional and add a txid filter

dcorral:
Adapt /listtransfers to the rgb-lib list_transfers redesign from RGB-Tools/rgb-lib#97: asset_id is now optional and an optional txid filter is added, with AND semantics when both are provided (at least one is required). Carries a temporary [patch.crates-io] pointing rgb-lib at git master until the next release is published on crates.io.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1174,6 +1174,48 @@ components:
           oneOf:
             - $ref: '#/components/schemas/Media'
             - type: 'null'
+    AssetFilter:
+      oneOf:
+        - $ref: '#/components/schemas/AssetFilterAnyOrNone'
+        - $ref: '#/components/schemas/AssetFilterNone'
+        - $ref: '#/components/schemas/AssetFilterId'
+      discriminator:
+        propertyName: type
+        mapping:
+          AnyOrNone: '#/components/schemas/AssetFilterAnyOrNone'
+          None: '#/components/schemas/AssetFilterNone'
+          Id: '#/components/schemas/AssetFilterId'
+    AssetFilterAnyOrNone:
+      type: object
+      required:
+        - type
+      properties:
+        type:
+          type: string
+          enum: [AnyOrNone]
+    AssetFilterId:
+      type: object
+      required:
+        - type
+        - value
+      properties:
+        type:
+          type: string
+          enum: [Id]
+        value:
+          type: string
+          example: rgb:CJkb4YZw-jRiz2sk-~PARPio-wtVYI1c-XAEYCqO-wTfvRZ8
+      example:
+        type: Id
+        value: rgb:CJkb4YZw-jRiz2sk-~PARPio-wtVYI1c-XAEYCqO-wTfvRZ8
+    AssetFilterNone:
+      type: object
+      required:
+        - type
+      properties:
+        type:
+          type: string
+          enum: [None]
     AssetIFA:
       type: object
       required:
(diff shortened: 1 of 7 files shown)


dcorral:
The 3 failing build (1.88.0, *) jobs are an MSRV clash, not related to the listtransfers change: the PR carries a temporary [patch.crates-io] pointing rgb-lib at git master (needed until the next release is on crates.io), and master's sea-orm 2.0 update bumped rgb-lib's MSRV to 1.94.0 while RLN still builds with 1.88.0. 

@zoedberg since RLN will hit this anyway when it picks up the next rgb-lib release, do you want me to bump RLN's rust-version and the CI matrix to 1.94.0 in this PR, or would you rather handle the MSRV bump separately?

zoedberg:
I think you can do it in this PR, but I'd first do a commit that sets master and bumps the MSRV, then the commit that updates the API

src/routes.rs
```diff
@@ -2858,8 +2859,18 @@ pub(crate) async fn list_transfers(
...
 
+    if payload.txid.is_none() && payload.asset_id.is_none() {
+        return Err(APIError::InvalidRequest(s!(
+            "either asset_id or txid must be provided"
+        )));
+    }
+    let asset_filter = match payload.asset_id {
+        Some(asset_id) => rgb_lib::wallet::AssetFilter::Id(asset_id),
+        None => rgb_lib::wallet::AssetFilter::AnyOrNone,
+    };
```
zoedberg: I think we should add a filter in RLN and convert it to the rgb-lib's `AssetFilter`, this way you can also drop the `either asset_id or txid must be provided` error

dcorral:
ready now for review @zoedberg

src/test/payment.rs
```diff
@@ -229,6 +229,43 @@ async fn success() {
...
+        by_txid.iter().map(|t| t.idx).collect::<Vec<i32>>(),
+        expected
+    );
+
+    let unknown_txid = "0000000000000000000000000000000000000000000000000000000000000000";
+    assert!(
+        list_transfers_by_asset_and_txid(node1_addr, &asset_id, unknown_txid)
+            .await
+            .is_empty()
+    );
```
zoedberg: this is testing rgb-lib's functionality, we already have tests in rgb-lib for this and RLN should just call its API without adding extra logic

openapi.yaml
```diff
@@ -2385,12 +2385,23 @@ components:
...
+            - string
+            - 'null'
           example: rgb:CJkb4YZw-jRiz2sk-~PARPio-wtVYI1c-XAEYCqO-wTfvRZ8
+        txid:
+          type:
+            - string
+            - 'null'
+          description: >-
+            Return the transfers committed by the on-chain transaction with this
+            txid. Combined with asset_id it acts as an intersection.
```
zoedberg: please no description on fields to keep the style consistent

zoedberg (changes requested):
Thanks! Please address the requested changes and rebase on top of the updated master tip

src/routes.rs
```diff
@@ -326,6 +326,24 @@ impl From<RgbLibAssetUDA> for AssetUDA {
     }
 }
 
+#[derive(Debug, PartialEq, Deserialize, Serialize)]
+#[serde(tag = "type", content = "value")]
+pub(crate) enum AssetFilter {
```
zoedberg: please respect the alphabetical order

zoedberg (changes requested):

zoedberg:
@dcorral please see the requested change and rename the second commit

src/routes.rs
```diff
@@ -180,6 +180,24 @@ impl From<RgbLibAssetCFA> for AssetCFA {
     }
 }
 
+#[derive(Debug, PartialEq, Deserialize, Serialize)]
```
zoedberg, line 183: Is `PartialEq` needed here? I don't see `AssetFilter` being compared anywhere; it looks like it's only serialized/deserialized and converted into `RgbLibAssetFilter`

zoedberg (changes requested):

zoedberg (approved):
LGTM, thanks

PR #152, opened 2026-08-10
Add BOLT11 description and description_hash to invoice creation, decode and payment APIs

dcorral:


openapi.yaml
```diff
@@ -2444,6 +2454,18 @@ components:
             - integer
             - 'null'
           example: 42
+        description:
+          type:
+            - string
+            - 'null'
+          description: Optional. When provided, the invoice includes a BOLT11 description. Mutually exclusive with description_hash.
```
zoedberg: usual description removal request ;)

src/ldk.rs
```diff
@@ -206,6 +208,9 @@ impl_writeable_tlv_based!(PaymentInfo, {
     (10, updated_at, required),
     (12, payee_pubkey, required),
     (14, expires_at, option),
+    // odd types so older binaries skip the fields instead of failing the whole read
```
zoedberg:
```suggestion
```

src/ldk.rs
```diff
@@ -206,6 +208,9 @@ impl_writeable_tlv_based!(PaymentInfo, {
...
+    (15, description, option),
+    (17, description_hash, option),
```
zoedberg: no need for odd numbers, use even ones

zoedberg (changes requested):
thanks, I did a partial review, please address the requested changes

openapi.yaml
```diff
@@ -2490,6 +2454,16 @@ components:
             - integer
             - 'null'
           example: 42
+        description:
```
zoedberg, line 2459: I would document that description and description_hash are mutually exclusive in the `/lninvoice` description

src/utils.rs
```diff
@@ -316,6 +337,28 @@ where
         .expect("request task panicked, see the preceding panic for the cause")
 }
 
+pub(crate) fn parse_invoice_description(
```
zoedberg: let's avoid methods that are called only once and aren't likely needed elsewhere soon. here `parse_invoice_description` adds 10+ lines of boilerplate, makes the mutual exclusivity of description/description_hash harder to see at the API boundary, and doesn't buy much, `ln_invoice` is short enough to hold the match inline

src/utils.rs
```diff
@@ -316,6 +337,28 @@ where
...
+    description_hash: Option<&str>,
+) -> Result<Bolt11InvoiceDescription, APIError> {
+    match (description, description_hash) {
+        (Some(_), Some(_)) => Err(APIError::InvalidRequest(s!(
+            "cannot provide both description and description_hash"
+        ))),
+        (Some(description), None) => Ok(Bolt11InvoiceDescription::Direct(
+            Description::new(description.to_string())
+                .map_err(|e| APIError::InvalidDescription(e.to_string()))?,
+        )),
```
zoedberg: I would treat `Some("")` as `None`

src/test/invoice.rs
```diff
@@ -94,11 +228,14 @@ async fn zero_amount_invoice() {
 
     // Create a zero-amount invoice on node2
     println!("Creating zero-amount invoice on node {node2_addr}");
+    let description = "1 cup of coffee";
```
zoedberg: I don't think it makes sense to test description in a zero-amount invoice test. Also you already add 2 tests for successful description and description_hash, you could test list_payments there

src/test/invoice.rs
```diff
@@ -196,11 +345,14 @@ async fn zero_amount_invoice() {
     assert_eq!(decoded_without_amount.asset_amount, None);
 
     // If the RGB invoice already includes asset_id and asset_amount, sendpayment can omit both.
+    let description_hash = lightning_invoice::Sha256(Sha256::hash(b"out-of-band description"));
```
zoedberg: same as above

src/utils.rs
```diff
@@ -262,6 +266,23 @@ pub(crate) async fn do_connect_peer(
...
+    }
+}
+
+pub(crate) fn description_hash_from_invoice(invoice: &Bolt11Invoice) -> Option<[u8; 32]> {
+    match invoice.description() {
+        Bolt11InvoiceDescriptionRef::Hash(hash) => Some(hash.0.to_byte_array()),
+        _ => None,
+    }
+}
+
```
zoedberg: I would do a `fn description_fields(invoice: &Bolt11Invoice) -> (Option<String>, Option<[u8; 32]>)` instead of these 2 methods

zoedberg (changes requested):

zoedberg (approved):
LGTM, thank you!

